PR #14, opened 2015-01-06
Classify PCI Sub-Class for better performance

yangoliver:
Current implementation just treats all serial bus devices as legacy
devices, and use IRQ_LEGACY by default. But in fact, some serial
bus controller like USB, Fiber Channel, InfiniBand are high speed
and high interrupt rate devices, which are similar with ethernet or
storage devices. On my server with 6 Fiber Channel devices, all these
devices interrupts could be bound to single CPUs under heavy work
load(24Gbps Throughput). That caused serious performance issue. But
if we classify all Fiber channel devices as storage class, which uses
CORE_LEVEL by default, it could fix the problem by balance interrupt
load among multiple cores.

This fix uses PCI sub-class code to classify all serial bus devices,
instead of just using major PCI class code.

Here are the major changes,

a. For USB devices, uses IRQ_OTHER
b. For Fiber Channel devices, uses IRQ_STORAGE
c. For InfiniBand devices, uses IRQ_STORAGE
d. For IPMI devices, uses IRQ_OTHER
e. All other devices, keep original classifications: IRQ_LEGACY

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -25,33 +25,6 @@ char *classes[] = {
 static int map_class_to_level[8] =
 { BALANCE_PACKAGE, BALANCE_CACHE, BALANCE_CORE, BALANCE_CORE, BALANCE_CORE, BALANCE_CORE, BALANCE_CORE, BALANCE_CORE };
 
-
-#define MAX_CLASS 0x12
-/*
- * Class codes lifted from pci spec, appendix D.
- * and mapped to irqbalance types here
- */
-static short class_codes[MAX_CLASS] = {
-	IRQ_OTHER,
-	IRQ_SCSI,
-	IRQ_ETH,
-	IRQ_VIDEO,
-	IRQ_OTHER,
-	IRQ_OTHER,
-	IRQ_LEGACY,
-	IRQ_OTHER,
-	IRQ_OTHER,
-	IRQ_LEGACY,
-	IRQ_OTHER,
-	IRQ_OTHER,
-	IRQ_LEGACY,
-	IRQ_ETH,
-	IRQ_SCSI,
-	IRQ_OTHER,
-	IRQ_OTHER,
-	IRQ_OTHER,
-};
-
 struct user_irq_policy {
 	int ban;
 	int level;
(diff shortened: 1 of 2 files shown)


yangoliver:
Hi Neil,

This bug fix could improve performance on a big server with multiple Fiber Channel cards.
In my test environment, 6 Fiber Channel ports got classified as IRQ_LEGACY devices. But each of them have 3 MSI-X IRQs, and the interrupt rate are quite similar with 10G Ethernet cards.

Let me know if you have any comments about this patch.

ppwaskie:
This patch looks reasonable to me.  I can pull it onto my box with some
10GbE NICs in them, but I don't think from visual inspection that they will
be broken with this.

yangoliver:
2015-01-06 14:51 GMT+08:00 PJ Waskiewicz <notifications@github.com>:
>
> This patch looks reasonable to me. I can pull it onto my box with some
> 10GbE NICs in them, but I don't think from visual inspection that they will
> be broken with this.

Yes.

I had verified the fix on a server with Fiber Channel Cards, USB
devices, 10G and 1G NIC cards.
There are no classification impacts to 10G and 1G NIC cards.

Here is my Fiber Channel card information,

 # lspci -s 0000:5d:00.0 -x
5d:00.0 Fibre Channel: QLogic Corp. Device 2831 (rev 02)
00: 77 10 31 28 46 01 10 00 02 00 04 0c 10 00 80 00
10: 0c a0 c0 ff 3f 38 00 00 0c 40 c0 ff 3f 38 00 00
20: 0c 00 b0 ff 3f 38 00 00 00 00 00 00 20 11 35 05
30: 00 00 d4 c7 44 00 00 00 00 00 00 00 0b 01 00 00

You can see the PCI class code information is  0x0c 04 00

a. base class 0xc for serial bus controller
b. sub class 0x04 for fiber channel
c. interface is 0x00

In original code, all 0xc major class is legacy.

But per this fix, as sub class is 0x04, it will be classified as
storage(CORE_LEVEL).


Below is my unit test output by gdb, you can see my fiber channel
device is classified as CORE_LEVEL,

Breakpoint 2, add_one_irq_to_db (devpath=0x7fffc70482b0
"/sys/bus/pci/devices/0000:5d:00.0", irq=332, pol=0x7fffc7048290)

    at classify.c:235

235     in classify.c

(gdb) p *new

$10 = {irq = 330, class = 2, type = 0, level = 3, flags = 0, numa_node
= 0x0, cpumask = {bits = {0 <repeats 64 times>}},

  affinity_hint = {bits = {0 <repeats 64 times>}}, hint_policy = 0,
irq_count = 0, last_irq_count = 0, load = 0, moved = 0,

  assigned_obj = 0x0}

Below is the output of --oneshot, my device is classified as storage.
Without the fix, it was legacy,

# /irqbalance --debug --oneshot | grep 330

Adding IRQ 330 to database

                  Interrupt 330 node_num is 1 (storage/0)

> On Mon, Jan 5, 2015 at 9:19 PM, Oliver <notifications@github.com> wrote:
>
> > Hi Neil,
> >
> > This bug fix could improve performance on a big server with multiple Fiber
> > Channel cards.
> > In my test environment, 6 Fiber Channel ports got classified as IRQ_LEGACY
> > devices. But each of them have 3 MSI-X IRQs, and the interrupt rate are
> > quite similar with 10G Ethernet cards.
> >
> > Let me know if you have any comments about this patch.
> >
> > --
> > Reply to this email directly or view it on GitHub
> > <https://github.com/Irqbalance/irqbalance/pull/14#issuecomment-68827396>.
> >
>
> --
> Reply to this email directly or view it on GitHub.




-- 
------------------
Oliver Yang

ppwaskie:
Awesome, looks good then to me.

Neil, want to merge or shall I?

-PJ

yangoliver:
I've updated my pull request to fix some typo issues in patch comments.
There is no C code change.

yangoliver:
Here is some background information about this bug.

Below are the maximum line speeds of major serial bus devices today, 

a.	Fiber Channel is 16Gbps.
b.	InfiniBand is 25Gbps.
c.	USB 3.0 is 5.0Gbps

On a server with multiple Fiber channel or InfiniBand cards, this irqbalance bug could cause very serious performance issue.

My testing was based on multiple 8GB Fiber channel cards.  I believe it would be more serious problem on 16GB Fiber channel cards.

In old irqbalance code (0.5x), as device classification was based on interrupt vector name, it could work well. After we upgrade irqbalance from old to new(v1.0.x), we ran into this problem.

yangoliver:
Neil,

I create a incremental commit for your code review.

I can merge them into one commit once you decide to pull my request.

nhorman:
No need to merge hem, this looks good.  Thanks!

PR #15, opened 2015-01-06
Fix sort_irqs()

zengjason:
Sort_irqs() should sort irq_info object by class firstly, then by load,
and finally by address. Current code considers class/load/address at the
same priority. This is not desirable because objects with lower class or
load value (but larger address) will be put before objects with larger
class or load value.

Signed-off-by: Jason Zeng <zrzeng@gmail.com>

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -614,7 +614,15 @@ static gint sort_irqs(gconstpointer A, gconstpointer B)
         a = (struct irq_info*)A;
         b = (struct irq_info*)B;
 
-	if (a->class < b->class || a->load < b->load || a < b)
+	if (a->class < b->class)
+		return 1;
+	if (a->class > b->class)
+		return -1;
+	if (a->load < b->load)
+		return 1;
+	if (a->load > b->load)
+		return -1;
+	if (a < b)
 		return 1;
         return -1;
 }


ppwaskie:
Hi Jason,

Can you outline a scenario that this is directly addressing?

zengjason:
Hi,

Before we calculate the irq placement, we need to sort the irqs, those irqs with higher workload should be in the front of the sorted list, so that when we pick up irqs one by one to calculate their placement, irqs with larger workload can be separated to different CPUs for example. That is to say, the order in the sorted list is important. Current code sorts the irq list in a wrong way.

As an example, suppose we have 4 CPUs, and 5 irqs. Suppose their information is as follows:

irq0: workload=2000, address=11111;
irq1: workload=2000, address=22222;
irq2: workload=2000, address=33333;
irq3: workload=1000, address=44444;
irq4: workload=1000, address=55555;

With current code, the sorted list is:
    irq3, irq4, irq0, irq1, irq2
so the final irq placement is:
    cpu0: irq3, irq2,
    cpu1: irq4, 
    cpu2: irq0,
    cpu3: irq1,

We can see that the result is not optimal. The workload of cpu1 is only 1000, while cpu0 is 3000.

With the fix, the sorted list would be:
    irq0, irq1, irq2, irq3, irq4
and the final irq placement is:
    cpu0: irq0
    cpu1: irq1,
    cpu2: irq2,
    cpu3: irq3, irq4

Each cpu has a workload of 2000.

nhorman:
I agree with pholasek, do you have any testing results that show this change has any benefit.  Initially that comparison was done because higher class codes consistently were shown to have higher irq volume than lower class codes.  While its possible for a scsi hba to have more interrupts than a 10G ethernet adapter, its really not likely, so I'd like to see some real world test results before we make a change here.

nhorman:
No response, in 2 months, closing.  Please reopen if you want to pursue this

PR #17, opened 2015-01-22
bug-893478_warn-once

posophe:
Only warn once for affinity hint subset empty irqs
https://bugzilla.novell.com/show_bug.cgi?id=893478


carnil:
Related Bug in Debian: https://bugs.debian.org/784391

PR #18, opened 2015-02-11
Support PCI quirks in classify code

yangoliver:
Current irqbalance uses PCI class code to determine the irq class
for each of devices. This logic could work well on most of devices.

However, the problem is, there are some PCI devices don't follow
the PCI spec. That caused the improper irq class code assigned to
these devices. Usually, it is hard to ask HW/FW vendors to fix this
kind of bug. Some of those are not bugs, because some new PCI class
code were not defined before the new devices shipment. At same time,
it is very difficult to force a device firmware upgrade even if the
new firmware fix is available for users.

As a correct irq class is crucial for interrupt balance efficiency
and performance, we need some software workarounds for these devices.
This is also what Linux kernel PCI code does in practice.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -5,6 +5,7 @@
 #include <sys/types.h>
 #include <dirent.h>
 #include <assert.h>
+#include <errno.h>
 
 #include "irqbalance.h"
 #include "types.h"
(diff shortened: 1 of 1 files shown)


ppwaskie:
This looks reasonable to me, nice work.

Neil, you want to apply it or shall I?

-PJ

yangoliver:
Neil,

Please let me know if you need more information about this fix.

The advantage of this patch is, we can easily extend our PCI quirks list if we found any other new faulty devices.

On my Linux box, I knew irq no 77 has wrong classification.  I used this irq as the test case.
With my new irqbalance fix, the device can be classified as storage device.

    # ./old_irqbalance --debug --oneshot | grep 77
    Adding IRQ 77 to database
              Interrupt 77 node_num is -1 (legacy/0)

    #./new_irqbalance --debug --oneshot | grep 77
    Adding IRQ 77 to database
                      Interrupt 77 node_num is -1 (storage/0)

yangoliver:
@Neil and @ppwaskie ,

I noticed that irqbalance had some new commits recently.
So I rebased my branch to make sure my code changes are based on latest HEAD.

Please review new updated pull request. 

There is no real code changes since last request. Only rebase operation.

nhorman:
I'm not opposed to this, it seems reasonable.  That said, I think if there is going to be any sort of volume to these quirks, it would be better to have apply_pci_quirks be a generic function that searches a table of data against the passed in irq device info, and rewrites the quirk based on that information.  I'll apply it as is now, but if you find a subsequent quirk, please modify that function to be table based.

yangoliver:
Neil,

Thanks for your comments.

I totally agree with you.about more generic code for table search. I will plan the improvements after I finished validations work on more different hardware.

PR #19, opened 2015-03-12
automatically add isolated cpus to banned_cpus

rikvanriel:
These patches add code to parse /proc/cmdline, extract the isolcpus= argument, and automatically add those isolated cpus to the banned_cpus mask.

This ensures that IRQs will not get assigned to CPUs that are supposed to be isolated from system activity.

### DIFF
--- a/bitmap.c
+++ b/bitmap.c
@@ -377,3 +377,87 @@ int __bitmap_parse(const char *buf, unsigned int buflen,
 
 	return 0;
 }
+
+/**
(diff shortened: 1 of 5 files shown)


jeremyeder:
I guess hasn't been encountered because of the level of hand-tuning that typically goes along with realtime.  Also realtime often does not use IRQ-based I/O.  Anyway, this is a cool option to have automated!

PR #20, opened 2015-03-23
Replace g_list_find with g_list_find_custom

yangoliver:
When irqbalance got configured with --without-glib2, the build
will break due to no g_list_find in local glib implementation.

Also the g_list_find just does the pointer match instead of content
match, the g_list_find_custom seems to be more reasonable.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -305,12 +305,20 @@ static int is_banned_irq(int irq)
 	return entry ? 1:0;
 }
 
+gint substr_find(gconstpointer a, gconstpointer b)
+{
+	if (strstr(b, a))
+		return 0;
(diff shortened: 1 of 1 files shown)


ppwaskie:
On Sun, Mar 22, 2015 at 10:52 PM, Oliver Yang <notifications@github.com> wrote:
> When irqbalance got configured with --without-glib2, the build
> will break due to no g_list_find in local glib implementation.
>
> Also the g_list_find just does the pointer match instead of content
> match, the g_list_find_custom seems to be more reasonable.

This looks good to me, nice catch.  Thanks for the patch.

Applied.

PR #22, opened 2015-05-05
find best place for larger workload ahead

K-DEN:
This makes a change in migration updating so that finding best place for rebalancing-required irqs with larger workload ahead of less ones. Eventual expected load variance becomes less.


nhorman:
Ok, I'll get around to changing that comment, thanks!

PR #23, opened 2015-05-22
Update and rename README to README.md

RoPP:


ppwaskie:
What's the purpose for the file rename?

RoPP:
This is just a cosmetic proposal. 
As github is the new official webpage, irqbalance could improve the readability of its readme using the markdown syntax : [updated readme preview](https://github.com/RoPP/irqbalance/blob/use-markdown-for-readme/README.md)

nhorman:
I suppose it doesn't hurt anything, Its still perfectly legible with vi after the change

Issue #33, opened 2016-12-21
Use new libsystemd library


bigon:
Hi,

Since a few version, libsystemd-journal has been merged in libsystemd itself.

The configure should be updated to detect this new library instead

PR #34, opened 2016-12-30
Search for libsystemd instead of libsystemd-journal

bigon:
libsystemd-journal has been merged in libsystemd at version 209
(2014-02-20), make the configure search for that "new" library.

Closes: #33

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -35,7 +35,10 @@ AC_ARG_WITH([systemd],
 )
 AS_IF(
   [test "x$with_systemd" = xyes], [
-    PKG_CHECK_MODULES([SYSTEMD], [libsystemd-journal], [journal_lib=yes])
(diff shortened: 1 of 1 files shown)


ppwaskie:
While this looks ok overall, have you tried this on an older system with an older version of systemd?  I'm more concerned about systems that want to upgrade irqbalance, but keep other things like systemd at an older level.

bigon:
It will indeed not work with the older versions of systemd. I'll see if I can add the detection of the older library as a fallback

ppwaskie:
Thanks.  I would be much more comfortable merging it if you can add that please.

-PJ

bigon:
OK, I've update my commit and force pushed it

ppwaskie:
Looks good, thanks Laurent!

PR #39, opened 2017-02-05
Add CI

szepeviktor:
Add Travis continuous integration to make sure new commits compile.

Signed-off-by: Viktor Szépe <viktor@szepe.net>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: c
+dist: trusty
+
+compiler:
+  - clang
+  - gcc
+
+script: ./autogen.sh && ./configure && make
--- a/ui/helpers.c
+++ b/ui/helpers.c
@@ -75,7 +75,8 @@ char * hex_to_bitmap(char hex_digit) {
 
 	char *bitmap = malloc(5 * sizeof(char));
 	bitmap[4] = '\0';
-	for(int i = 3; i >= 0; i--) {
+	int i;
+	for(i = 3; i >= 0; i--) {
 		bitmap[i] = digit % 2 ? '1' : '0';
 		digit /= 2;
 	}


ppwaskie:
I cannot take this as-is at all.  Please redo the commit and pull request with a proper checkin comment as to what you're doing in the commit, along with a proper Signed-off-by.  Also address the issue from szepeviktor.

szepeviktor:
> along with a proper Signed-off-by. Also address the issue with the automake you referenced..

Excuse me. I have no GPG set up and I do not speak C.

szepeviktor:
The *checkin comment* could be:

> Add Travis continuous integration to make sure new commits compile.

I am missing tests from this project.

szepeviktor:
> -s, --signoff
>    Add Signed-off-by line by the committer at the end of the commit log message.

I didn't know it is just a string, no GPG involved.

Although I am not able to properly fix the automake issue.

nhorman:
move the declaration of int i out of the for loop and put it at the top of the function.  Though I'm not sure what version of gcc you are using that disallows that sort of declaration.  Has travisCI enabled some pedantic checking of some sort?

szepeviktor:
@nhorman Could you restart Travis?

nhorman:
@szepeviktor it restarts itself, and it failed again. It appears you are getting clang failures, which isn't suprising as I don't think we've ever built irqbalance with clang

nhorman:
actually, scratch that, I'm confused here.  It appears the clang build ran before the gcc build and the former passed while the latter failed,  So I do need to restart travis, but I was under the impression it would autostart on new branch commits.  Looking at the settings, it appears that you should get an automated build start if your travis.yml file in place, which you do, so it should be restarting the next time you push your branch

nhorman:
Actually, are you just asking me to pull this?  Looking at it, your two commits are fine, and it would enable travis for everyone.

szepeviktor:
image

It seem like Travis is disabled. It does not show up in the bow at the bottom.

PR #42, opened 2017-03-01
Prefer pkgconfig to check for ncursesw.

Polynomial-C:
That way we take possible separate tinfo lib into account.

Otherwise the following happens when ncurses was built with separate tinfo library:
```
libtool: link: x86_64-pc-linux-gnu-gcc -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -march=native -mtune=native -O2 -pipe -Wl,-O1 -Wl,--hash-style=gnu -Wl,--sort-common -o irqbalance-ui ui/helpers.o ui/irqbalance-ui.o ui/ui.o  -Wl,--as-needed -lglib-2.0 -lcurses -lm -lnuma
/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/../../../../x86_64-pc-linux-gnu/bin/ld: ui/ui.o: undefined reference to symbol 'nodelay'
/lib64/libtinfo.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:564: irqbalance-ui] Error 1
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,7 +26,13 @@ AC_CHECK_FUNCS(getopt_long)
 AC_CHECK_LIB(numa, numa_available)
 AC_CHECK_LIB(m, floor)
 
-AC_CHECK_LIB(curses, mvprintw)
+PKG_CHECK_MODULES([NCURSESW], [ncursesw], [has_ncursesw=yes], [AC_CHECK_LIB(curses, mvprintw)])
+AS_IF([test "x$has_ncursesw" = "xyes"], [
+  AC_SUBST([NCURSESW_CFLAGS])
+  AC_SUBST([NCURSESW_LIBS])
+  LIBS="$LIBS $NCURSESW_LIBS"
+  AC_SUBST([LIBS])
+])
 
 AC_C_CONST
 AC_C_INLINE


ppwaskie:
Is this still an issue?  I'm unable to reproduce this.

Polynomial-C:
Yes it is still an issue on my Gentoo systems.

ppwaskie:
I need a bit more info on this then.  Almost all of my systems here are running Gentoo on x86_64 (laptop, 2 single-socket systems, and one multi-socket system).  I don't see the issue you're running into.  I don't have ncurses built with anything different though, just straight out of portage.  I can't speak to custom builds.  I'd need to do deeper testing on other distros to make sure this build change doesn't muck up standard stuff to accommodate custom configs.

What I'm running:

[I] sys-libs/ncurses
     Available versions:  
     (5)    5.9-r101(5/5)
     (0)    6.0-r1(0/6)
       {ada +cxx debug doc gpm minimal profile static-libs test threads tinfo trace unicode ABI_MIPS="n32 n64 o32" ABI_PPC="32 64" ABI_S390="32 64" ABI_X86="32 64 x32"}
     Installed versions:  5.9-r99(5/5)(03:16:59 PM 09/05/2015)(cxx gpm unicode -ada -static-libs -tinfo ABI_MIPS="-n32 -n64 -o32" ABI_PPC="-32 -64" ABI_S390="-32 -64" ABI_X86="32 64 -x32") 6.0-r1(01:57:26 AM 07/27/2017)(cxx gpm unicode -ada -debug -doc -minimal -profile -static-libs -test -threads -tinfo -trace ABI_MIPS="-n32 -n64 -o32" ABI_PPC="-32 -64" ABI_S390="-32 -64" ABI_X86="32 64 -x32")
     Homepage:            https://www.gnu.org/software/ncurses/ http://dickey.his.com/ncurses/
     Description:         console display library

Polynomial-C:
You only run into this issue when you enable the `tinfo` USE flag for ncurses. Please be careful when doing this as you will likely break most ncurses consumers with that change.
I have that flag enabled to catch packages that fail to find separate tinfo lib. With `sys-libs/ncurses[tinfo]` I can reproduce this on all my Gentoo systems (stabe, unstabe, amd64, x86).

ppwaskie:
I've run this through a number of distros, including my Gentoo systems.  It appears to be fine.  Applying.

Issue #43, opened 2017-03-02
Conflicting C89 and C99 Mode Requirement


rkubes:
A clean build of the master branch is failing on both the latest version of Raspbian on a Raspberry Pi, and Ubuntu 14.04 on a PC.

If left as-is, the build will fail on the UI code, with the following error:
```
ui/helpers.c: In function ‘hex_to_bitmap’:
ui/helpers.c:78:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for(int i = 3; i >= 0; i--) {
  ^
ui/helpers.c:78:2: note: use option -std=c99 or -std=gnu99 to compile your code
```
If this error is resolved manually, additional errors remain which also indicate the requirement C99 mode. All of these errors occur in the UI code. (*See edit below)

If the Makefile is modified to use -std=c99, the build will fail on the non-UI code with the following error:
```
numa.c:45:2: error: initializer element is not constant
  .mask = CPU_MASK_ALL,
  ^
numa.c:45:2: error: (near initialization for ‘unspecified_node_template.mask’)
```

The build will succeed if I build the UI portion on C99 mode, but the rest on C89 mode; but the Makefile does not seem to be configured to do this automatically.

Edit:
Upon further review, I realized all of the UI errors that indicate the requirement of C99 mode are the same. There are a number of 'for' loops with initial declarations. The build will be successful if all of the declarations are taken out of the 'for' loops.

The other remaining concerns are just warnings:
```
ui/irqbalance-ui.c: In function ‘main’:
ui/irqbalance-ui.c:378:6: warning: ignoring return value of ‘fgets’, declared with attribute warn_unused_result [-Wunused-result]
      (void)fgets(cmdstring, 255, f);
      ^
ui/irqbalance-ui.c:403:1: warning: control reaches end of non-void function [-Wreturn-type]
 }
 ^
```

ppwaskie:
Closing as this builds fine with gcc 5.x and beyond.  Please upgrade your compilers.

PR #44, opened 2017-03-11
Fix broken build

necheffa:
This pull request is with respect to the issue reported here:
https://github.com/Irqbalance/irqbalance/issues/43

There is some preprocessor magic that requires --std=c90 or older to build however there were some for loops using C99 syntax [i.e for(int i = 0; ...)]. This pull request switches these C99 style for loops to syntax compatible with older C standards, allowing irqbalance to build using the default CFLAGS generated by Autotools.

### DIFF
--- a/ui/helpers.c
+++ b/ui/helpers.c
@@ -75,7 +75,7 @@ char * hex_to_bitmap(char hex_digit) {
 
 	char *bitmap = malloc(5 * sizeof(char));
 	bitmap[4] = '\0';
-    int i = 3;
+	int i = 0;
 	for(i = 3; i >= 0; i--) {
 		bitmap[i] = digit % 2 ? '1' : '0';
 		digit /= 2;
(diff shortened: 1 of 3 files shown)


necheffa:
I just updated to GCC 5.4.0 (from 4.9.4) and irqbalance was able to build as-is without applying these changes. So I'm guessing the issue comes down to how the preprocessor directives behave in relation to GCC version used.

ppwaskie:
Is this still an issue?  I'm not seeing any problems with the build, and am considering dropping this PR.

necheffa:
Technically the build still fails on GCC 4.9 but given that GCC 7.1 was released a couple months ago and to my knowledge the stable release of all major distros is at least running 5.x branches of GCC; it seems best to just say "stop using an old compiler" and reject this merge request/close the associated issue (#4).

ppwaskie:
I'm going to go ahead and close this then.  I've been able to compile this all the way back to gcc-5.0.1.

PR #49, opened 2017-07-11
Teach irqbalance about Intel CoD

kmjohansen:
The Intel Cluster on Die topology breaks an existing assumption in irqbalance about the structure of the cpu and memory topology.  In CoD, it's possible for a physical package to have multiple NUMA domains.  In that case, irqbalance incorrectly constructs the topology in a way that strands one of the NUMA nodes.  This patch attempts to fix the problem by allowing NUMA nodes to point to the first portion of the CPU hierarchy where they are the unique.  At least with systems that the submitter has tested on, this turns out to be the LLC.

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -121,8 +121,32 @@ out:
 	log(TO_CONSOLE, LOG_INFO, "Banned CPUs: %s\n", buffer);
 }
 
-static struct topo_obj* add_cache_domain_to_package(struct topo_obj *cache, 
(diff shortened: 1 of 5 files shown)


ppwaskie:
This looks reasonable to me.  Neil?

ppwaskie:
I looked at this a few more times and ran a few tests on different setups I have here with varying NUMA topologies.  This looks solid.  I'm going ahead and merging this.

PR #51, opened 2017-07-17
Fix assigning IRQ affinity on the wrong NUMA node

### LINKED ISSUE #50: Incorrectly balance IRQ on system w/ multiple NUMA node in a CPU package
"irqbalance -d" incorrectly reports NUMA information on system w/ multiple NUMA nodes in a CPU package. For example, on AMD EPYC 2-socket server platform, which has 4 NUMA node per socket, irqbalance reports the following: 

> irqbalance --foreground -d | grep Package
> Package 0:  numa_node is 1 cpu mask is ffffffff,00000000,ffffffff (load 0)
> Package 1:  numa_node is 5 cpu mask is ffffffff,00000000,ffffffff,00000000 (load 0) 

This results in assigning IRQ affinity on the wrong NUMA node, and causing I/O performance issue.

szepeviktor:
props Thomas Lendacky

Closes #50

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -324,6 +324,8 @@ static void do_one_cpu(char *path)
 
 	nodeid=-1;
 	if (numa_avail) {
+		struct topo_obj *node;
+
 		dir = opendir(path);
 		do {
 			entry = readdir(dir);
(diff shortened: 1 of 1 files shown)


ppwaskie:
This looks reasonable to me.

PR #58, opened 2017-08-22
Fix compile warning and spell error

wanghaibin-mouse:


procinterrupts.c
```diff
@@ -57,7 +57,17 @@ static int check_platform_device(char *name, struct irq_info *info)
 	DIR *dirfd;
 	char path[512];
 	struct dirent *ent;
-	int rc = -ENOENT;
+	int rc = -ENOENT, i;
+	static struct pdev_irq_info{
```
ppwaskie: Space after pdev_irq_info and {

nhorman: concur with @ppwaskie , fix the space error and we can pull this

PR #67, opened 2018-06-07
Service file improvements

lucaswerkmeister:
See individual commit messages.

---

The sandbox is not perfect: if an attacker gains arbitrary code execution in `irqbalance`, they can still, for example, read any file owned by root¹ (`/etc/shadow`, `/etc/ssh/ssh_host_rsa_key`, etc.) and encode it inside files in `/proc/irq`², where another unprivileged process also controlled by the attacker could read it and publish it over the network. But I think the unit file would become too unwieldy if we added too many more directives – this should already provide a lot of protection at little cost.

¹: Without `CAP_DAC_READ_SEARCH`, it can’t read files that are owned by other users and not world-readable (e. g. stuff in `/home`), but files owned by root can be read without any special capabilities.
²: The kernel won’t let you create arbitrary files in `/proc/irq` (`ENOENT`) nor write arbitrary text to existing files there (`EIO`), but you could still send messages by encoding them over time, byte by byte, in the affinities for various IRQs.

---

I’m currently running a more aggressive version of this sandbox (https://github.com/lucaswerkmeister/server-etc/commit/38f3b87d5b4d48ed98c96980ea3dfc959f2a12bc) on my server, and so far it seems to work fine.

### DIFF
--- a/misc/irqbalance.service
+++ b/misc/irqbalance.service
@@ -7,6 +7,12 @@ ConditionVirtualization=!container
 [Service]
 EnvironmentFile=/path/to/irqbalance.env
 ExecStart=/usr/sbin/irqbalance --foreground $IRQBALANCE_ARGS
+CapabilityBoundingSet=
+NoNewPrivileges=yes
+ReadOnlyPaths=/
+ReadWritePaths=/proc/irq
+RestrictAddressFamilies=AF_UNIX
+TemporaryFileSystem=/run:ro
 
 [Install]
 WantedBy=multi-user.target
--- a/misc/irqbalance.service
+++ b/misc/irqbalance.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=irqbalance daemon
+Documentation=man:irqbalance(1)
+Documentation=https://github.com/Irqbalance/irqbalance
 ConditionVirtualization=!container
 
 [Service]


ppwaskie:
Can you please a Signed-off-by: line to the commit message?

I see the reasoning here, but if an attacker gains access in general and can get root access, then they could probably circumvent this as well.  Still thinking about it.

lucaswerkmeister:
Signed-off-by added, and I also added a few more directives because I remembered systemd/systemd#8846.

lucaswerkmeister:
You can test the sandbox interactively, by the way:

```sh
sudo systemd-run -qt \
  -p CapabilityBoundingSet= \
  -p NoNewPrivileges=yes \
  -p ReadOnlyPaths=/ \
  -p ReadWritePaths=/proc/irq \
  -p PrivateNetwork=yes \
  -p TemporaryFileSystem=/run:ro \
  $SHELL
```

nhorman:
Have you tried the irqbalance-ui interface with this changeset?  Setting privatenetwork I imagine might prevent access from that utility

lucaswerkmeister:
I wasn’t aware of that program (Debian doesn’t even include it in its `irqbalance` package, apparently). Trying it out on Arch, it looks like it’s not working (though I’m not sure, as `irqbalance-ui` doesn’t seem to display any errors – but I can see `ECONNREFUSED` errors from `connect(…, @"irqbalance….sock")` system calls in `strace`). With `PrivateNetwork` disabled, `irqbalance` dies with `SIGABRT` instead (buffer overflow, core dump), but I have no idea if that’s caused by the sandbox or just a general bug in irqbalance.

nhorman:
Yeah, it appears to be (or at least its suggested).  It works for me with PrivateNetwork disabled, and doesn't connect with it enabled.  Not sure why you're getting the SIGABRT, thats not happening for me at the head of the tree

nhorman:
Ping, any update here?

nhorman:
ok, closing due to lack of response. Please reopen if you would like to fix this up.

lucaswerkmeister:
Well, I was going to come back to this on the weekend, so should I reopen right now (I would like to fix it up) or only tomorrow? :)

nhorman:
ok, then I'll reopen it, but I closed it because you didn't respond to my requests for intent for the last 3 weeks

lucaswerkmeister:
Yes, I let this rot in my inbox too long, sorry :( I’ve pushed a new version with `RestrictAddressFamilies` instead of `PrivateNetwork`, can you perhaps try it out? On my system, `irqbalance-ui`’s first `connect()` now succeeds, but after that `irqbalance` dies with “buffer overflow detected”.

lucaswerkmeister:
Nevermind, if I actually install and run `irqbalance` from git instead of trying to run `irqbalance-ui` from master against `irqbalance` from latest release, it works :) I guess there’s been some incompatibility in the IPC protocol since the last release?

nhorman:
There shouldn't have been, but I suppose its possible, its pretty young code.  I'll look into that shortly, but yes, it works now for me as well.  Thanks for taking the extra time on it, i'll pull this

lucaswerkmeister:
Cool, thanks! And sorry again for dropping the ball there.

nhorman:
no worries!  you got it squared up, thats what counts!

PR #76, opened 2018-08-30
Fix several memleak problems found by covscan

ryncsn:
Some memleak issues is found by static analysis tools, and can confirm
irqbalance is leaking memory slowly when there are incomming connection
to socket.

This patch could solve the memleak problem.


nhorman:
Thanks, I never should have pulled the ui commit.  This all looks good to me.

PR #77, opened 2018-09-02
Fix irqbalance ui failing to connect to irqbalance daemon

ryncsn:
irqbalance ui is faling due to the changes in commit 19c25dd.
This patch align irqbalance-ui's socket connecting routine with
irqbalance.c


ppwaskie:
Crap, that's my fault.  I should have caught that.  Thanks.

PR #78, opened 2018-09-02
Fix an possible overflow error

ryncsn:
Got:
"specified bound 2048 exceeds the size 19 of the destination"
when -O2 is used, and a "*** buffer overflow detected ***" error output
with no backtrace.

With -O0, it's gone, guess it's some gcc optimization problem, and the
size there is wrong anyway, this patch could fix it.


ppwaskie:
This seems reasonable.  What version of gcc, just curious?

ryncsn:
`gcc --version`
`gcc (Gentoo Hardened 7.3.0-r3 p1.4) 7.3.0`

Thanks for merging

PR #79, opened 2018-10-11
policyscript: ignore non-executable files when specifying a directory

ryncsn:
Currenly irqbalance require custom scirpts to be executable. If a
non-executable file is found, irqbalance will stil try to execute it but
fail with no permission error.

So just ignore non-executable file, this also make is possible to have
non-executable files in policyscript directory as reference.


ryncsn:
Thanks for the review, I've updated the pull request.

nhorman:
Looks good, thanks!

PR #84, opened 2018-11-09
Update document and help message

ryncsn:


ryncsn:
Hi, thanks for the correction, I've updated this PR, and actually I copy&pasted some document from --banirq part, so I fixed an existing typo btw in the update.

nhorman:
Ack, looks good.  Thanks.

PR #86, opened 2018-11-30
systemd: add prefixed with "-" to EnviromentFile.

kometchtech:
- When failing to read the env file, do not judge it as a failure.
- https://www.freedesktop.org/software/systemd/man/systemd.exec.html

It is intended to prevent errors from forgetting to fix the file path as it is.
This is because we think that importing EnvironmetFile is not mandatory.

### DIFF
--- a/misc/irqbalance.service
+++ b/misc/irqbalance.service
@@ -5,7 +5,7 @@ Documentation=https://github.com/Irqbalance/irqbalance
 ConditionVirtualization=!container
 
 [Service]
-EnvironmentFile=/path/to/irqbalance.env
+EnvironmentFile=-/path/to/irqbalance.env
 ExecStart=/usr/sbin/irqbalance --foreground $IRQBALANCE_ARGS
 CapabilityBoundingSet=
 NoNewPrivileges=yes


ppwaskie:
Looks good to me, thanks for the submission.

kometchtech:
I am sorry that my words are missing.

As you say, I think it is a place to check the settings so that the introducer will not forget it.

However, reading the env file is not important. The user who executes up to the env file by specifying the environment variable thinks less about forgetting about the setting of this place.

It would be only the large meaning of insurance as my intention.

nhorman:
If I understand your words correctly, you are implying that you want to suppress the File Not Found error because you want the default case, in which no environment file is provided to 'just work'.  Is that correct?

If so, I disagree with that assertion.  While I understand the desire to have it all just work, I feel it important to force a system integrator to make a decision at install time about an evirionment file location.  If we want to ignore the error I would prefer just to remove the EnvironmentFile line from the unit file entirely, rather than have an intentionally non-existant path fail silently.

kometchtech:
Yes, your first half words are what I meant.

I understand your words about the second half words.
As you say, rather than using easy error control, the implementer should understand the behavior and set it.
I also agree with this.

I withdraw this pull request.

Sorry for inconvenience.

ppwaskie:
I'll revert it, sorry Neil. It seemed reasonable to me given the systemd
documentation. But I see your point

nhorman:
Its fine, no need to apologize, I just think its important to ensure that users:

a) have a clear hook point to add environmental settings and command line options
b) be forced to make a conscious decision about what they want in terms of environment

If you all can think of a less intrusive way to do that, I'm happy to discuss it and make changes, what we have here is just the best I could think of when we wrote it

PR #90, opened 2019-02-26
Refine document about IRQBALANCE_BANNED_CPUS

ryncsn:
There is no declaration about how irqbalance deal with isolated CPUs or
adaptive-ticks CPUs, and how IRQBALANCE_BANNED_CPUS will override the
behavior of auto-banning of such CPUs. Refine the document to avoid
confusion.

Signed-off-by: Kairui Song <kasong@redhat.com>

### DIFF
--- a/irqbalance.1
+++ b/irqbalance.1
@@ -155,6 +155,8 @@ Same as --debug.
 .TP
 .B IRQBALANCE_BANNED_CPUS
 Provides a mask of CPUs which irqbalance should ignore and never assign interrupts to.
+If not specified, irqbalance use mask of isolated and adaptive-ticks CPUs on the
+system as the default value.
 
 .SH "SIGNALS"
 .TP


nhorman:
I appreciate the patch, but the text isn't quite accurate.  the banned cpu mask is or-ed with the isolated and adaptive tick cpu sets.  It would be more accurate to leave out the text about the banned cpu set being unspecified and say instead that isolated and adaptive tick cpus are always considered to be banned.

ryncsn:
Thanks for the review, but I did get what you mean by:
> the banned cpu mask is or-ed with the isolated and adaptive tick cpu sets

The doc text is based on the commit message of ca5a3f13bf247561e64ef2323b3dd28a2c3b880d and 2dddcfef7f17e8df197513b21c077294b4487263, they says that use of IRQBALANCE_BANNED_CPUS enviroment variable will disable the auto detect of cpu masks, and that's how the function setup_banned_cpus() works.

For example I isolated cpu 0-3:
Auto-detect worked well:
```
[root@localhost ~]# irqbalance --debug
Isolated CPUs: 0000000f
Adaptive-ticks CPUs: 00000000
Banned CPUs: 0000000f
```

IRQBALANCE_BANNED_CPUS did disable the auto-detection:
```
[root@localhost ~]# export IRQBALANCE_BANNED_CPUS="0000010"
[root@localhost ~]# irqbalance --debug
Isolated CPUs: 00000000
Adaptive-ticks CPUs: 00000000
Banned CPUs: 00000010
```

I could post another patch to fix the behavior instead.

nhorman:
Oh yo're right, I missed the goto out in that commit.  This is ok

PR #92, opened 2019-03-25
Remove a duplicated word in manpage

ryncsn:
Just noticed this waning in a manpage scan report, trivial but let's
fix it.

Signed-off-by: Kairui Song <kasong@redhat.com>


ppwaskie:
Looks fine to me, thanks for the fix.

PR #98, opened 2019-06-06
should consider user and sys load when account cpu's load

Pyxisha:
currently, only irq and sortirq are considered when account cpu's load.
we find two defect in practice.

1. CONFIG_IRQ_TIME_ACCOUNTING may be not configured in some system's
kernel .config. if CONFIG_IRQ_TIME_ACCOUNTING is not set, irq time will
always be zero. and may cause some irq has thousands of interrupts but
irq's load is 1 just like this:
        Cache domain 0:  numa_node is 1 cpu mask is
		01000000,00000000,00000000  (load 0)
                CPU number 88  numa_node is 1 (load 0)
                  Interrupt 468 node_num is 1 (ethernet/1:0)
                  Interrupt 1006 node_num is -1 (storage/1:0)
                  Interrupt 1026 node_num is -1 (storage/1:0)
                  Interrupt 999 node_num is -1 (storage/1:278733)
                  Interrupt 1011 node_num is -1 (storage/1:0)
                  Interrupt 1013 node_num is -1 (storage/1:0)
                  Interrupt 986 node_num is -1 (storage/1:0)
                  Interrupt 1007 node_num is -1 (storage/1:0)
                  Interrupt 1027 node_num is -1 (storage/1:0)
                  Interrupt 1010 node_num is -1 (storage/1:216261)
                  Interrupt 1012 node_num is -1 (storage/1:0)
                  Interrupt 985 node_num is -1 (storage/1:0)
2. some process may be bonded to a specific cpu.

Signoff_by: XiaShuang <xiashuang1@huawei.com>

### DIFF
--- a/procinterrupts.c
+++ b/procinterrupts.c
@@ -434,7 +434,7 @@ void parse_proc_stat(void)
 	size_t size = 0;
 	int cpunr, rc, cpucount;
 	struct topo_obj *cpu;
-	unsigned long long irq_load, softirq_load;
+	unsigned long long user_load, sys_load, irq_load, softirq_load;
 
 	file = fopen("/proc/stat", "r");
 	if (!file) {
@@ -463,8 +463,8 @@ void parse_proc_stat(void)
 		if (cpu_isset(cpunr, banned_cpus))
 			continue;
 
-		rc = sscanf(line, "%*s %*u %*u %*u %*u %*u %llu %llu", &irq_load, &softirq_load);
-		if (rc < 2)
+		rc = sscanf(line, "%*s %llu %*u %llu %*u %*u %llu %llu", &user_load, &sys_load, &irq_load, &softirq_load);
+		if (rc < 4)
 			break;	
 
 		cpu = find_cpu_core(cpunr);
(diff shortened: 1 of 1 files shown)


nhorman:
I understand the issue you are seeing, but I'm not sure this solution makes sense.  If CONFIG_IRQ_TIME_ACCOUNTING isn't configured in a kernel, then irqbalance cant really do its job really doing its job.  Just balancing irqs based on user and system time for a set of cpus will cause irqbalance to just continually move irqs away from cpus that are doing heavy application work.  On the surface that sounds ok, but irqbalance is meant to distribute irq/softirq load as evenly as possible accross all eligible cpus, not avoid cpus that have heavy userspace load.  The schedulers job is to move processes away from heavily loaded cpus.  I think the better solution is to require CONFIG_IRQ_TIME_ACCOUNTING on the running kernel, but scanning /proc/stat on startup, and if all irq_time values in /proc/stat are one, exit with an error indicating that CONFIG_IRQ_TIME_ACCOUNTING isn't configured

Pyxisha:
I see what you mean. In practice, over half of systems in Huawei are not configed CONFIG_IRQ_TIME_ACCOUNTING for performance concern. Is there a solution that can let irqbalance works in these systems? @nhorman 
best wishes.

nhorman:
Off the top of my head, no, I'm afraid.  The purpose of irqbalance is to distribute the load generated by irqbalance evenly among cpus, so as to level througput for each cpu.  Without the ability to determine what that load generated by irqs is, irqbalance can't do its job.   My best suggestion to you would be, in the event that you don't have IRQ_TIME configured, would be to use the --policyscript option to pass a policy script in which attempts to use some other information to spread interrupts out.  you could develop a policy script that simply attempts to do a distribution of irqs equally to cpus or some such.

PR #101, opened 2019-07-04
Clarify the README description

paride:


paride:
Maybe it's me, but without knowing what irqbalance actually does the description may be interpreted wrongly, as in: irqbalance isolates *all* the highest irq sources to a single CPU.

PR #102, opened 2019-07-05
a hole that may flees hotplug event.

Pyxisha:
in parse_proc_interrupts, irqbalance will set the global var need_rescan to 1 if a hot plug event occured. and in scan, if need_rescan == 1, it will clear the irq db and rescan the interrupts:
		parse_proc_interrupts();
		parse_proc_stat();
		sleep_approx(sleep_interval);
		clear_work_stats();
		parse_proc_interrupts();
a probalble issue is: hot pulg event may occured again, and
need_rescan may be set to 1 again. in such situation, interrupts need to be
rescaned but not. so there may be a hole that flees hotplug event.

I think a possible solution is return TRUE instead of going down
directly in the need_rescan branch.

Xiashuang <xiashuang1@huawei.com>

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -270,9 +270,9 @@ gboolean scan(gpointer data)
 		for_each_irq(NULL, force_rebalance_irq, NULL);
 		parse_proc_interrupts();
 		parse_proc_stat();
-		sleep_approx(sleep_interval);
-		clear_work_stats();
-		parse_proc_interrupts();
+
+		/* Still need to check hotplugged or not next round */
+		return TRUE;
 	}
 
 	parse_proc_stat();


nhorman:
I don't see why this is necessary, parse_proc_interrupts will set need_rescan based on (among other conditions), if the cpu count from /proc/interrupts doesn't match the last value recorded (via core_count), so if we get a hotplug event while rebuilding the irq database, we will just not affine interrupts to that cpu until the next time the scan loop runs (10 seconds later by default).  All this patch does is guarantee that scan will run twice in rapid succession without clearing the old work stats in between.  I don't think it hurts anything too significant, but it doesn't really solve any problems either.  Do you have a real example of a problem here?

Pyxisha:
In some scenarios， hotplug events may occured continuously. suppose such a situation:
1. irq1, irq2, irq3 are added.
2. parse_proc_interrupts set need_rescan to 1
3. parse_proc_interrupts
4. irq4, irq5,  irq6 are added
5. sleep 10s
6. parse_proc_interrupts again and need_rescan are set to 1 agin
7. irq1, irq2 and irq3 are balanced but irq4, irq5and irq are not.
8. sleep 10s
9. 2. parse_proc_interrupts set need_rescan to 1

In such a scenario, irq4, irq5 and irq6 may not be balanced in 20 seconds. this may cause interrupts centralize in a specific core in the period of time,  and performance reduction. in some ultra scenario， may even cause the system can’t touch watchdog and panic. Actually, this patch have not other side effects， so， I think this patch is a acceptable solution.

Pyxisha:
In some scenarios， hotplug events may occured continuously. suppose such a situation:

1. irq1, irq2, irq3 are added.
2. parse_proc_interrupts set need_rescan to 1
3. irq4, irq5, irq6 are added
4. parse_proc_interrupts, need_rescan are setg to 1 again
5. sleep 10s
6. parse_proc_interrupts
7. need_rescan is 1 but not rescan. irq1, irq2 and irq3 are balanced but irq4, irq5 and irq6 are not.
8. sleep 10s
9. parse_proc_interrupts
10. rescan, and irq4 and ir5 and irq6 are  balanced.
In such a scenario, irq4, irq5 and irq6 may not be balanced in 30 seconds. this may cause interrupts centralize in a specific core in the period of time, and result in performance reduction. in some ultra scenario， may even cause the system can’t touch watchdog and panic. This patch will make irq4, irq5 irq6 being balanced faster in the previous example. This patch actually have not other side effects， so， I think this patch is a acceptable solution.

nhorman:
Your scenario is invalid.

because step (3) happens prior to the second calling of parse_proc_interrupts in step (4), then the rebuild of the irq database will include all interrupts listed (1,2,3,4,5, and 6), via the call to build_object_tree, and you're find

The worst case scenario is if you add irqs/cpus between steps (4) and (5).  That will delay the balancing of irqs for the 10 second delay to gather delta stats, and another 10 seconds until the next call of scan, but they will get balanced.

You're concern over an inability to touch the watchdog timer is also completely invalid.  By default newly added irqs are affined by the kernel to all cpus, not the null set of cpus. 

If you really think you have found a bug which causes some adverse effect, show me a reproducer.  currently your thought experiment is invalid.

Pyxisha:
I'm don't think this is a bug.  what I mean is that in the need_rescan branch, hotpulg event may happened  after build_object_tree() and before sleep_approx(sleep_interval), so in line 272 of irqbalance.c, need_rescan will be 1 but irqbalance flees the hotpulg event to next round. I think return True here is a better solution, hotplug event may be processed in advance， and  duplicated code can be reduced.

As for "newly added irqs are affined by the kernel to all cpus", Yeah， it is true. but some drivers may set smp_affinity to a default set of CPU, and by default, all interrupts will centralize in the first cpu of the default set.

PR #106, opened 2019-07-11
use policy prior to the default values

Pyxisha:
Currently user-defined policies against non-PCI devices' interrupts
are not working properly.

For example, when trying to set "balance_level=core" for a non-PCI
device interrupt which is classified as "other", will result in
the level of "package" because overrided in add_new_irq().

This patch fixes this by restricting irq info initializations in
add_one_irq_to_db(), which requires a change on its parameters.

Signed-off-by: Shuang Xia <xiashuang1@huawei.com>
Signed-off-by: Yun Wu <wuyun.wu@huawei.com>

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -343,10 +343,10 @@ void add_cl_banned_module(char *modname)
  * related device. NULL devpath means no sysfs entries for
  * this irq.
  */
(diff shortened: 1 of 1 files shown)


nhorman:
yep, that looks good thanks

PR #107, opened 2019-07-12
after Irqbalance@dcc411e, the activate_mapping function seems a little redundancy, 

Pyxisha:
the variable valid_mask and some judgement seems unnecessary.

sign_off_by: Shuang Xia <xiashuang1@huawei.com>

### DIFF
--- a/activate.c
+++ b/activate.c
@@ -60,8 +60,6 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 {
 	char buf[PATH_MAX];
 	FILE *file;
-	cpumask_t applied_mask;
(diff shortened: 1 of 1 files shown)


nhorman:
This looks reasonable.  Since you're working in that function, you can further reduce this.  check_affinity really just compares the assigned mask to the one currently set in proc.  You can remove the applied_mask parameter from check_affinity, and use info->assigned_obj->mask instead, saving you a parameter to pass to that function.  That also allows you to remove the applied_mask stack variable from activate_mapping as well

Pyxisha:
Yeah.

PR #109, opened 2019-07-15
fix sleep interval when sleep_interval is changed by socket

Pyxisha:
currently, in scan, irqbalance compare sleep_interval's address to decide if sleep_interval is changed, accutually this judgement is always false now.

sign_off_by: Shuang Xia <xiashuang1@huawei.com>

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -61,6 +61,7 @@ char *pidfile = NULL;
 char *polscript = NULL;
 long HZ;
 int sleep_interval = SLEEP_INTERVAL;
+int last_interval;
 GMainLoop *main_loop;
 
 char *banned_cpumask_from_ui = NULL;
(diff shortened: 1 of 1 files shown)


nhorman:
why are you using last_interval in main?  That shouldn't be needed

Pyxisha:
Yeah, I foo. mislead by the original code.

nhorman:
Have you tested this to ensure that changing the sleep time via the irqbalance ui to ensure that the scan interval updates properly?

Pyxisha:
I add a print here:
gboolean scan(gpointer data __attribute__((unused)))
{
        struct timespec now;
        clock_gettime(CLOCK_MONOTONIC, &now);

        log(TO_CONSOLE, LOG_INFO, "%d\n\n\n-----------------------------------------------------------------------------\n", now.tv_sec);

[debug.txt](https://github.com/Irqbalance/irqbalance/files/3397100/debug.txt)

and from the debug output, We can see the sleep interval change from 10s to 5s.

PR #110, opened 2019-07-17
The manpage belongs to section 8

paride:


nhorman:
irqbalance is also functional on sysv variants, which has no section 8.  How is this going to work with those older systems?

paride:
TIL about the SysV man sections. I guess it's a good reason to dismiss this PR. Thanks for the prompt feedback!

PR #111, opened 2019-07-21
Add stub manpage for irqbalance-ui

paride:


nhorman:
I think you want to make the inclusion of the man page conditional on the IRQBALANCEUI autoconf variable, otherwise, you'll get a man page installed without the corresponding binary

paride:
Done. If you're wondering where these contributions are coming from, thing is I'm adopting the `irqbalance` Debian package, and I'm cleaning up some lintian warning :)

nhorman:
thanks!

PR #116, opened 2019-09-10
change path to devpath

liuchao173:
keep the parameters 'path' of check_for_irq_ban function consistent with line 699.
In check_for_irq_ban,
sprintf(cmd, "%s %s %d > /dev/null",banscript, path, irq);
the banscript is unique, so i think the path should keep consistent.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -599,7 +599,7 @@ static int check_for_module_ban(char *name)
 		return 0;
 }
 
-static int check_for_irq_ban(char *path __attribute__((unused)), int irq, GList *proc_interrupts)
+static int check_for_irq_ban(int irq, GList *proc_interrupts)
 {
 	struct irq_info find, *res;
 	GList *entry;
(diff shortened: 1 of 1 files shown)


nhorman:
that path variable is actually unused, so it should just be removed entirely, not made consistent.

liuchao173:
It's unused in irqbalance, but the path and irq are user banscript's parameters

nhorman:
you're missing my point.  check_for_irq_ban actually no longer uses the passed in path parameter, and so the path parameter can be removed from that function call.  So you don't need to make the call points use simmilar variables, you can just eliminate that parameter from the list and all the call points entirely

nhorman:
why did you close this?  I asked you to make some improvements.  Are you unwilling to make changes?

liuchao173:
you mean remove the path in check_for_irq_ban? I misunderstand

nhorman:
yes, that exactly.

PR #117, opened 2019-09-11
Initialize blen

liuchao173:
Recommend initializing blen to 0, since cpustr is NULL and it's alloced in getline.


ppwaskie:
Does this fix anything, like a compiler warning, or a unit test?  To me this seems like a very harmless stack variable that gets wiped out anyways later.  So why trouble the compiler to deal with an assignment?

liuchao173:
It fixs nothing, just to keep the code style consistent. If the lineptr is alloced in getline, the parameter n is usually initialized to 0.

PR #119, opened 2019-09-19
make the return value of getline() handled correct

yeyunfeng-dev:
getline() will return -1 when fail, so make the return value handle
correct.


ppwaskie:
Looks good to me.  Thank you.

PR #120, opened 2019-09-19
make the parameter line of getline be freeed correctly

hejingxian123:
1 When getline ERRORNO is ENOMEM, the input line will return null. we can't free line when it's null.
2 This buffer should be freed by the user program even if getline() failed.
3 Threre exists file op with no closing before return.

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -278,13 +278,13 @@ static void do_one_cpu(char *path)
 	if (file) {
 		char *line = NULL;
 		size_t size = 0;
-		if (getline(&line, &size, file)<=0)
-			return;
-		fclose(file);
-		if (line && line[0]=='0') {
+		if ((getline(&line, &size, file) <= 0) ||
+			(line && line[0] == '0')) {
+			fclose(file);
 			free(line);
 			return;
 		}
+		fclose(file);
 		free(line);
 	}
 


nhorman:
Theres a simpler way to do this.  POSIX defines free as allowing the freeing of NULL (reducing to a noop).  As such, instead of checking the value of line for being NULL subsequent to the getline call, just set line = NULL prior to the getline call.  Then you just unconditionally free it in the error case

hejingxian123:
The commit has been simpler.

nhorman:
The setup_banned_cpu changes look good, but the do_one_cpu changes could still be simplified.  See the review I started

hejingxian123:
I can't get your meaning  about do_one_cpu. I has reverted the changes about do_one_cpu. You can accept the changes about setup_banned_cpu firstly. Thanks.

nhorman:
What don't you understand?  I provided you a sample diff.

hejingxian123:
I have made changes abount do_one_cpu more simpier. But  I haven't find your provide diff file, due to   I have no experience in this.  If you have more suggestion, you can describe it in the comment. Thanks very much.

nhorman:
Its literally 4 comments up in this discussion

hejingxian123:
I think my modification is already the simplest way to avoid resource release.  I have seen the above comments  more than one thousand times. You say 'Theres a simpler way to do this. POSIX defines free as allowing the freeing of NULL (reducing to a noop). As such, instead of checking the value of line for being NULL subsequent to the getline call, just set line = NULL prior to the getline call. '. I have already done the same as you say. Do you see my modification again? There is already no problem in my modification.  This time I only add ten lines, and it's better than the exist code in irqbalance.

nhorman:
I don't know how to make it any clearer for you.  If you don't want to take the suggestions I've offered, I'll just take care of it myself.

PR #121, opened 2019-09-21
delete unused local variables in parse_proc_interrupts(), and checking return value of strdup() in collect_full_irq_list()

yeyunfeng-dev:
The local variable @savedline[1024] is unused in parse_proc_interrupts(),
so delete it.

### DIFF
--- a/procinterrupts.c
+++ b/procinterrupts.c
@@ -189,6 +189,8 @@ GList* collect_full_irq_list()
 			continue;
 
 		savedline = strdup(line);
+		if (!savedline)
+			break;
 		irq_name = strtok_r(savedline, " ", &savedptr);
 		if (strstr(irq_name, "xen-dyn") != NULL)
 			is_xen_dyn = 1;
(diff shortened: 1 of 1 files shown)


ppwaskie:
These seem fine to me.  Thanks for the fixes.

nhorman:
Woah!  I had an open review on this, and the changes as listed will imply that any transient low memory error will result in us not balancing an entire set of irqs

ppwaskie:
Hmm, sorry about that Neil.  I wasn't aware.  I thought it looked ok, but
if you see something I didn't, then please revert.

-PJ

yeyunfeng-dev:
Hi Neil,
This modify I reference the error handling of getline() in collect_full_irq_list():
      if (getline(&line, &size, file)<=0)
                         break;
so the patch use break too when strdup() fail. 

If any advice, I can modify for it. I think falling back to use a local array when strdup() fail is much better ?

yeyunfeng-dev:
On 2019/9/23 23:24, Neil Horman wrote:
> No worries, its not really worth reverting at this point, I'll just fix it up next time I'm in there. I think I have some pending changes to commit anyway.
> 

This modify I reference the error handling of getline() in collect_full_irq_list():
	if (getline(&line, &size, file)<=0)
		break;
so the patch use break too when strdup() fail.

If any advice, I can modify for it. I think falling back to use a local array when strdup() fail is much better ?

Thanks.

PR #123, opened 2019-10-06
free the memory when getline() fail in add_one_node()

yeyunfeng-dev:
when getline() fail, the memory still need to be freed.

### DIFF
--- a/numa.c
+++ b/numa.c
@@ -74,12 +74,11 @@ static void add_one_node(const char *nodename)
 		cpus_clear(new->mask);
 	} else {
 		ret = getline(&cpustr, &blen, f);
-		if (ret <= 0) {
+		if (ret <= 0)
 			cpus_clear(new->mask);
-		} else {
+		else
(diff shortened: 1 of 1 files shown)


numa.c
```diff
@@ -78,8 +78,8 @@ static void add_one_node(const char *nodename)
 			cpus_clear(new->mask);
 		} else {
 			cpumask_parse_user(cpustr, ret, new->mask);
-			free(cpustr);
 		}
```
nhorman: The movement of the free call looks good, but now that the else clause is now a single statement, could you please eliminate the braces  to keep it in line with the rest of the file style please?  In fact, can you fix the if clause too while you're at it?  I should have done that one long ago.

nhorman:
comments made in the patch

numa.c
yeyunfeng-dev: I have modify the patch as your suggest, thanks.

PR #124, opened 2019-10-10
prevent NULL pointer dereference when memory allocation failure

yeyunfeng-dev:
There are several places where memory allocation does not check return
values, adding null pointer checks.

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -313,6 +313,9 @@ void get_irq_data(struct irq_info *irq, void *data)
 	else
 		*irqdata = realloc(*irqdata, strlen(*irqdata) + 24 + 1 + 11 + 20 + 20 + 11);
 
+	if (!*irqdata)
+		return;
+
 	sprintf(*irqdata + strlen(*irqdata),
 			"IRQ %d LOAD %lu DIFF %lu CLASS %d ", irq->irq, irq->load,
 			(irq->irq_count - irq->last_irq_count), irq->class);
@@ -343,6 +346,9 @@ void get_object_stat(struct topo_obj *object, void *data)
 	} else {
 		*stats = realloc(*stats, strlen(*stats) + irqdlen + 31 + 11 + 20 + 11 + 1);
 	}
+
+	if (!*stats)
+		return;
 
 	sprintf(*stats + strlen(*stats), "TYPE %d NUMBER %d LOAD %lu SAVE_MODE %d %s",
 			object->obj_type, object->number, object->load,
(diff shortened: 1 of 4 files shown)


irqbalance.c
```diff
@@ -360,14 +366,16 @@ gboolean sock_handle(gint fd, GIOCondition condition, gpointer user_data __attri
...
 	struct iovec iov = { buff, 500 };
 	struct msghdr msg = { 0 };
 	msg.msg_iov = &iov;
 	msg.msg_iovlen = 1;
 	msg.msg_control = malloc(CMSG_SPACE(sizeof(struct ucred)));
 	msg.msg_controllen = CMSG_SPACE(sizeof(struct ucred));
 
-	struct cmsghdr *cmsg;
+	if (!msg.msg_control)
+		return FALSE;
```
nhorman: I don't think this is right, as it will cancel the watch on the socket.  You still want to accept the connection and receive the message, dropping it before you return.

yeyunfeng-dev:
On 2019/10/10 20:05, Neil Horman wrote:
> *@nhorman* commented on this pull request.
> 
> ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
> 
> In irqbalance.c <https://github.com/Irqbalance/irqbalance/pull/124#discussion_r333480500>:
> 
>>  	struct iovec iov = { buff, 500 };
>  	struct msghdr msg = { 0 };
>  	msg.msg_iov = &iov;
>  	msg.msg_iovlen = 1;
>  	msg.msg_control = malloc(CMSG_SPACE(sizeof(struct ucred)));
>  	msg.msg_controllen = CMSG_SPACE(sizeof(struct ucred));
>  
> -	struct cmsghdr *cmsg;
> +	if (!msg.msg_control)
> +		return FALSE;
> 
> I don't think this is right, as it will cancel the watch on the socket. You still want to accept the connection and receive the message, dropping it before you return.
> 
so return TRUE instead of return FALSE ?
what is "dropping it before you return." mean ?

thanks.

irqbalance.c
yeyunfeng-dev: ok，I see. 
I have modify the patch by checking the cmsg instead of msg.msg_control. thanks.

nhorman:
That looks better, thank you.

I'll accept this along with the closedir fix (which also looks good), but please in the future, if you find other issues, open new pull requests for them.  Tacking on additional fixes that don't relate to the pull request topic will cause confusion when reading the history

PR #126, opened 2019-10-11
correct to use realloc() function

yeyunfeng-dev:
The man doc about realloc() say:
  "
  If realloc() fails the original block is left untouched; it is not
  freed or move
  "

So make the handling of realloc() function correctly.

In addition, there is another problem about parameter using in
sock_handle(), it should be use the address of @setup instead of @setup.

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -308,13 +308,17 @@ gboolean scan(gpointer data __attribute__((unused)))
 void get_irq_data(struct irq_info *irq, void *data)
 {
 	char **irqdata = (char **)data;
-	if (!*irqdata)
-		*irqdata = calloc(24 + 1 + 11 + 20 + 20 + 11, 1);
-	else
-		*irqdata = realloc(*irqdata, strlen(*irqdata) + 24 + 1 + 11 + 20 + 20 + 11);
+	char *newptr = NULL;
 
 	if (!*irqdata)
+		newptr = calloc(24 + 1 + 11 + 20 + 20 + 11, 1);
+	else
+		newptr = realloc(*irqdata, strlen(*irqdata) + 24 + 1 + 11 + 20 + 20 + 11);
+
+	if (!newptr)
 		return;
+
+	*irqdata = newptr;
 
 	sprintf(*irqdata + strlen(*irqdata),
 			"IRQ %d LOAD %lu DIFF %lu CLASS %d ", irq->irq, irq->load,
@@ -325,6 +329,7 @@ void get_object_stat(struct topo_obj *object, void *data)
 {
 	char **stats = (char **)data;	
 	char *irq_data = NULL;
+	char *newptr = NULL;
 	size_t irqdlen;
 
 	if (g_list_length(object->interrupts) > 0) {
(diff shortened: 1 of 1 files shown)


irqbalance.c
```diff
@@ -308,13 +308,20 @@ gboolean scan(gpointer data __attribute__((unused)))
...
+		old = *irqdata;
+		*irqdata = realloc(old, strlen(old) + 24 + 1 + 11 + 20 + 20 + 11);
+	}
 
-	if (!*irqdata)
+	if (!*irqdata) {
+		/* if realloc fail, not free the original memory */
+		*irqdata = old;
 		return;
+	}
```
nhorman: You can make this far simpler. Just declare a new void pointer and make that store the return value of calloc/realloc, then check that, return if NULL, and in the fall through case, assign the result to *irqdata

 void get_irq_data(struct irq_info *irq, void *data)
 {
        char **irqdata = (char **)data;
+       void *newptr;
        if (!*irqdata)
-               *irqdata = calloc(24 + 1 + 11 + 20 + 20 + 11, 1);
+               newptr = calloc(24 + 1 + 11 + 20 + 20 + 11, 1);
        else
-               *irqdata = realloc(*irqdata, strlen(*irqdata) + 24 + 1 + 11 + 20 + 20 + 11);
+               newptr = realloc(*irqdata, strlen(*irqdata) + 24 + 1 + 11 + 20 + 20 + 11);
 
+       if (!newptr)
+               return;
+       *irqdata = newptr;
+

irqbalance.c
```diff
@@ -344,11 +352,16 @@ void get_object_stat(struct topo_obj *object, void *data)
...
+		*stats = realloc(old, strlen(old) + irqdlen + 31 + 11 + 20 + 11 + 1);
 	}
 
-	if (!*stats)
+	if (!*stats) {
+		/* if realloc fail, not free the original memory */
+		*stats = old;
+		free(irq_data);
 		return;
+	}
```
nhorman: same as the above comment, use a temp variable to capture the return of calloc/realloc, test it for null, and return if it is, or assign it to *stats otherwise

irqbalance.c
```diff
@@ -465,17 +478,21 @@ gboolean sock_handle(gint fd, GIOCondition condition, gpointer user_data __attri
...
-				for_each_irq(cl_banned_irqs, get_irq_data, setup);
+				for_each_irq(cl_banned_irqs, get_irq_data, &setup);
 			}
 			cpumask_scnprintf(banned, 512, banned_cpus);
-			setup = realloc(setup, strlen(setup) + strlen(banned) + 7 + 1);
-			if (!setup)
+			old = setup;
+			setup = realloc(old, strlen(old) + strlen(banned) + 7 + 1);
+			if (!setup) {
+				free(old);
```
nhorman: Instead of freeing here, assign the temp variable to the result of realloc, and, if its _not_ null, assign it to *setup, then call sprintf, and send.  That way you only have a single call to free at the bottom of the function

irqbalance.c
yeyunfeng-dev: ok, I have modify the patch. thanks.

irqbalance.c
yeyunfeng-dev: ok, I have modify the patch. thanks.

irqbalance.c
yeyunfeng-dev: ok, I have modify the patch. thanks.

PR #129, opened 2019-10-20
fix resource leak on the error paths in main()

yeyunfeng-dev:

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -547,6 +547,7 @@ int init_socket()
 int main(int argc, char** argv)
 {
 	sigset_t sigset, old_sigset;
+	int ret = EXIT_SUCCESS;
 
 	sigemptyset(&sigset);
 	sigaddset(&sigset,SIGINT);
@@ -637,7 +638,7 @@ int main(int argc, char** argv)
 			    "single cpu.  Shutting down\n";
 
 		log(TO_ALL, LOG_WARNING, "%s", msg);
-		exit(EXIT_SUCCESS);
+		goto out;
 	}
 
 
@@ -660,7 +661,8 @@ int main(int argc, char** argv)
 	parse_proc_stat();
 
 	if (init_socket()) {
-		return EXIT_FAILURE;
+		ret = EXIT_FAILURE;
+		goto out;
 	}
 	main_loop = g_main_loop_new(NULL, FALSE);
 	last_interval = sleep_interval;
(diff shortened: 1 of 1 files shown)


irqbalance.c
```diff
@@ -530,13 +530,17 @@ int init_socket()
 		addr.sun_family = AF_UNIX;
 		if (bind(socket_fd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
 			log(TO_ALL, LOG_WARNING, "Daemon couldn't be bound to the abstract socket, bailing out.\n");
+			close(socket_fd);
+			socket_fd = 0;
```
nhorman: theres nothing wrong with this, but you don't need to rewrite socket_fd to 0 on failure, as any subsequent call will just rewrite it again in the new call to init_socket.

Also, please recode this using a label pattern (like you did in the main routine), so that we don't have multiple close calls in this function.  It would be better to have a ret variable that gets set to 0 or 1 with a goto and a label at the function epiloge. Something like:
ret = 0
...
if(socket_fd < 0) {
    ret = 1
   goto fail
}
....
done:
     return ret
fail:
     close(socket_fd)
     goto done

yeyunfeng-dev: because socket_fd will be close in main(), so rewrite socket_fd to 0. so do not close socket_fd in init_socket() ?
thanks.

yeyunfeng-dev: I have update the patch, not close socket_fd in init_socket(), thanks.

nhorman: this is better, yes.  But if you go this route, there is another location in main (the check for core count < 2), where you can goto the fail label and do the same free operaion.

yeyunfeng-dev: ok, I will modify as your suggest.

yeyunfeng-dev: I have update the patch, thanks.

irqbalance.c
```diff
@@ -637,7 +638,8 @@ int main(int argc, char** argv)
 			    "single cpu.  Shutting down\n";
 
 		log(TO_ALL, LOG_WARNING, "%s", msg);
-		exit(EXIT_SUCCESS);
+		ret = EXIT_SUCCESS;
```
nhorman: you don't need this here, ret is already assigned EXIT_SUCCESS in the prologue

yeyunfeng-dev: ok, thanks.

PR #136, opened 2019-10-30
irqbalance: prevent NULL pointer dereference in do_one_cpu()

yeyunfeng-dev:
add_cpu_to_cache_domain() will return NULL if memory allocation failure,
so check the return value before invoking add_cache_domain_to_package().

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -176,8 +176,10 @@ static struct topo_obj* add_cache_domain_to_package(struct topo_obj *cache,
 
 	if (!entry) {
 		package = calloc(1, sizeof(struct topo_obj));
-		if (!package)
+		if (!package) {
+			need_rebuild = 1;
 			return NULL;
+		}
 		package->mask = package_mask;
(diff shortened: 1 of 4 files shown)


yeyunfeng-dev:
I have update the patch as your suggest, thanks.

nhorman:
Ok, thank you, that looks like the approach I was thinking of, but don't you want to extend the rebuild flag to get set for the various oom conditions that might arise throughout the build_object_tree process?  like if calloc fails in add_one_node or if the calloc fails in do_one_cpu directly?

yeyunfeng-dev:
ok, but can I  new commit to add  rebuild flag, becuase I have another pull request:  "move numa.c to cputree.c and refactor some numa codes"? thanks.

nhorman:
No, do it here, I've asked before that you not mix and match pull requests like that

yeyunfeng-dev:
ok, I have update the patch, thanks.

PR #137, opened 2019-10-31
move numa.c to cputree.c and refactor some numa codes

yeyunfeng-dev:
move numa.c to cputree.c
refactor the code about numa in cputree


nhorman:
NAK, We only happen to use numa info for cpu topology purposes at the moment, but numa describes much more than locality of cpus.  It also describes locality of memory and peripherals, which we may want to take advantage of one day.  I'd much rather leave numa code separate from cpu topology code

yeyunfeng-dev:
ok, thanks.

PR #138, opened 2019-11-03
use NUMA_NO_NODE macro and num_online_cpus() instead

yeyunfeng-dev:
use NUMA_NO_NODE macro instead of -1 for numa
use num_online_cpus() instead core_count

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -45,12 +45,11 @@ GList *cache_domains;
 GList *packages;
 
 int cache_domain_count;
-int core_count;
 
 /* Users want to be able to keep interrupts away from some cpus; store these in a cpumask_t */
 cpumask_t banned_cpus;
 
-cpumask_t cpu_possible_map;
+cpumask_t cpu_online_map;
 
 /* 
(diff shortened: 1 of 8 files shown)


cputree.c
```diff
@@ -45,12 +45,11 @@ GList *cache_domains;
 GList *packages;
 
 int cache_domain_count;
-int core_count;
 
 /* Users want to be able to keep interrupts away from some cpus; store these in a cpumask_t */
 cpumask_t banned_cpus;
 
-cpumask_t cpu_possible_map;
+cpumask_t cpu_online_map;
```
nhorman: if you're going to do this, you may as well remove the cpu_possible_map extern and its associated macros.

yeyunfeng-dev: cpu_present_map is also unused too, is it need to remove together ?
thanks.

nhorman: may as well, sure

yeyunfeng-dev: ok, patch has been updated now.

nhorman: looks good

PR #139, opened 2019-11-06
irqbalance: initialize package_mask before using in do_one_cpu()

yeyunfeng-dev:
The local variable package_mask is used without initialized. add
cpus_clear(package_mask) before using.

In addition, the variable cache_mask is called cpu_set() twice, which
is unnecessary, so delete the redundant code.

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -299,12 +299,13 @@ static void do_one_cpu(char *path)
 		return;
 	}
 
-	cpu_set(cpu->number, package_mask);
-
 	/* try to read the package mask; if it doesn't exist assume solitary */
 	snprintf(new_path, ADJ_SIZE(path, "/topology/core_siblings"),
 		 "%s/topology/core_siblings", path);
-	process_one_line(new_path, get_mask_from_bitmap, &package_mask);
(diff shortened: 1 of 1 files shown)


cputree.c
```diff
@@ -299,6 +299,7 @@ static void do_one_cpu(char *path)
 		return;
 	}
 
+	cpus_clear(package_mask);
```
nhorman: No, this is completely unnecessecary.  package_mask is cleared as part of cpumask_parse_user.  In commit c7122666e0f72634512284fe90258daf1dca5914 you should have just removed the call to cpu_set entirely when you converted to the use of the process_one_line function.  Please just remove the call to cpu_set

yeyunfeng-dev: The comment "try to read the package mask; if it doesn't exist assume solitary", so it means process_one_line() maybe will fail, it need a default package_mask. is that true? thanks.

nhorman: Yes, if process_one_line fails, you will then need to clear and set the mask in the error condition.  either way though, you don't need to clear it before hand.

yeyunfeng-dev: ok, I have update the patch, thanks.

PR #141, opened 2019-11-08
fix invalid pointer dereference banned_cpumask_from_ui

dublio:
The memory of cpu_ban_string was release in sock_handle function,
so the banned_cpumask_from_ui will dereference an invalid memory.

Fix this issue by delay release memory.

Reproduce:
echo "settings cpus 0-3" | nc -U `find /var/run/irqbalance/ -name *sock`

Signed-off-by: Weiping Zhang <zhangweiping@didiglobal.com>

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -39,6 +39,7 @@
 #include "irqbalance.h"
 
 extern char *banned_cpumask_from_ui;
+extern char *cpu_ban_string;
 
 GList *cpus;
 GList *cache_domains;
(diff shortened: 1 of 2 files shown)


nhorman:
yeah, this looks like a legitimate bug, but thats alot of change to fix the problem.  Why not just allocate cpu_ban_string in sock_handle, like we currently are, and have setup_banned_cpu check that pointer instead, do the grooming on the string (ie:
banned_cpumask_from_ui = strtok(cpu_ban_string, " ");
                                if (!strncmp(banned_cpumask_from_ui, "NULL", strlen("NULL"))) {
                                        banned_cpumask_from_ui = NULL;
                                }
In setup_banned_cpus, and free it unilaterally at the end?  That would just move 4 lines of code around and solve the problem just as well

dublio:
> yeah, this looks like a legitimate bug, but thats alot of change to fix the problem. Why not just allocate cpu_ban_string in sock_handle, like we currently are, and have setup_banned_cpu check that pointer instead, do the grooming on the string (ie:
> banned_cpumask_from_ui = strtok(cpu_ban_string, " ");
> if (!strncmp(banned_cpumask_from_ui, "NULL", strlen("NULL"))) {
> banned_cpumask_from_ui = NULL;
> }
> In setup_banned_cpus, and free it unilaterally at the end? That would just move 4 lines of code around and solve the problem just as well

Good idea ^_^

dublio:
@nhorman Hi, I have update patch with a bit more check before allocate memory for cpu_ban_string to avoid memory leak if prior one has not been released.

nhorman:
This looks better, thanks

PR #142, opened 2019-11-11
use process_one_line() instead of fscanf() and remove redundancy check in add_one_irq_to_db()

yeyunfeng-dev:
irqbalance: use process_one_line() instead of fscanf()
irqbalance: remove redundancy check in add_one_irq_to_db()

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -285,17 +285,6 @@ void add_cl_banned_irq(int irq)
 	add_banned_irq(irq, &cl_banned_irqs);
 }
 
-static int is_banned_irq(int irq)
-{
-	GList *entry;
-	struct irq_info find;
-
-	find.irq = irq;
-
-	entry = g_list_find_custom(banned_irqs, &find, compare_ints);
-	return entry ? 1:0;
-}
-
 gint substr_find(gconstpointer a, gconstpointer b)
 {
 	if (strstr(b, a))
@@ -340,21 +329,6 @@ static struct irq_info *add_one_irq_to_db(const char *devpath, struct irq_info *
 	struct irq_info *new;
 	int numa_node;
 	char path[PATH_MAX];
-	GList *entry;
-
-	/*
-	 * First check to make sure this isn't a duplicate entry
-	 */
-	entry = g_list_find_custom(interrupts_db, hint, compare_ints);
-	if (entry) {
-		log(TO_CONSOLE, LOG_INFO, "DROPPING DUPLICATE ENTRY FOR IRQ %d on path %s\n", irq, devpath);
-		return NULL;
-	}
-
-	if (is_banned_irq(irq)) {
-		log(TO_ALL, LOG_INFO, "SKIPPING BANNED IRQ %d\n", irq);
-		return NULL;
-	}
 
 	new = calloc(1, sizeof(struct irq_info));
 	if (!new)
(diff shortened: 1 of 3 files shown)


nhorman:
This doesn't seem right.  You're claiming that get_irq_info is a valid substitue for  checking to see if the irq is on the banned list, but its not.  If an irq has been added to the banned list, it won't ever be added to the irq db, and with this change, it will

yeyunfeng-dev:
get_irq_info() find irq_info from interrupts_db and banned_irqs, the logic is same as the check logic in add_one_irq_to_db()，isn't it ? thanks.

nhorman:
No, get_irq_info returns the irq_info struct for an interrupt, be it on the banned_list or the actual irq db list.  is_banned_irq tells you if the interrupt is on the banned list or not

yeyunfeng-dev:
I agree with you, but the check logic in add_one_irq_to_db() not only call is_banned_irq(), but also check is irq in interrupts_db list:
	/*
	 * First check to make sure this isn't a duplicate entry
	 */
	entry = g_list_find_custom(interrupts_db, hint, compare_ints);
	if (entry) {
		log(TO_CONSOLE, LOG_INFO, "DROPPING DUPLICATE ENTRY FOR IRQ %d on path %s\n", irq, devpath);
		return NULL;
	}

nhorman:
yes, which your commit removes.  See the add_new_irq path.  I think you're assuming that all callers of add_one_irq_to_db check the banned list, and thats not the case

yeyunfeng-dev:
`
static void add_new_irq(int irq, struct irq_info *hint, GList *proc_interrupts)
{
	struct irq_info *new;
	struct user_irq_policy pol;

	new = get_irq_info(irq);   //--> check irq whether is in interrupts_db  banned_irqs list or not
	if (new)
		return;

	/* Set NULL devpath for the irq has no sysfs entries */
	get_irq_user_policy(NULL, irq, &pol);
	if ((pol.ban == 1) || check_for_irq_ban(irq, proc_interrupts)) { /*FIXME*/
		add_banned_irq(irq, &banned_irqs);   //--> only add to the banned_irqs list
		new = get_irq_info(irq);
	} else
		new = add_one_irq_to_db(NULL, hint, &pol); //-->so it is unnecessary check again
`
look at the code add_new_irq(), I have make some comments, before invoking add_one_irq_to_db(), it has already been checked by get_irq_info().  Am I right? thanks.

nhorman:
you're still missing my point.  In add_new_irq, yes, you check to see if you have already added this irq, which checks both the irq_db and banned lists, thats fine.  but in the event you don't find an irq, you check your policy, and potentially add the irq to the banned list, then call add_one_irq_to_db.  In add_one_irq_to_db, you have removed the check for is_banned_irq, and so, in that situation you will wind up adding an irq that you have already put on the banned list to the irq db.  Dont do that.

yeyunfeng-dev:
the point I can't understand is that:
" you check your policy, and potentially add the irq to the banned list, then call add_one_irq_to_db"

if the irq add to  banned list, it will not call add_one_irq_to_db(), it is two different path code:
`
if ((pol.ban == 1) || check_for_irq_ban(irq, proc_interrupts)) { /*FIXME*/
	add_banned_irq(irq, &banned_irqs); 
	new = get_irq_info(irq);
} else
	new = add_one_irq_to_db(NULL, hint, &pol); 
`
thanks.

nhorman:
Oh, my bad, you're right, I missed the else clause there.

PR #143, opened 2019-12-02
free polscript if irqbalance exit

liuchao173:
polscript is alloced in parse_command_line and isn't freed when irqbalance exit


nhorman:
You don't need the conditional in there.  free() does the right thing when you pass NULL in.  You also want to fix up the parse function, to free the old polscript if the same option is passed twice on the comand line

PR #150, opened 2020-03-11
arm64: Add irq aff change check

liuchao173:
For aarch64, the PPIs format in /proc/interrputs can be parsed and add to interrupt db, and next, the number of interrupts is counted and used to calculate the load. Finally these interrupts maybe scheduled between the NUMA domains.

Acctually, the PPIs cannot change aff, and it should not be added to interrupt db. This patch fix it.

Add a check before add a interrupt to db, just only reads the irq's aff, and write it back to avoid any impact on the system, According to the result of writing to fitler the irq.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -256,7 +256,7 @@ static gint compare_ints(gconstpointer a, gconstpointer b)
 	return ai->irq - bi->irq;
 }
 
-static void add_banned_irq(int irq, GList **list)
+static void __add_banned_irq(int irq, GList **list)
 {
 	struct irq_info find, *new;
 	GList *entry;
@@ -280,9 +280,14 @@ static void add_banned_irq(int irq, GList **list)
 	return;
 }
 
+void add_banned_irq(int irq)
+{
+	__add_banned_irq(irq, &banned_irqs);
+}
+
 void add_cl_banned_irq(int irq)
 {
-	add_banned_irq(irq, &cl_banned_irqs);
+	__add_banned_irq(irq, &cl_banned_irqs);
 }
 
 gint substr_find(gconstpointer a, gconstpointer b)
@@ -374,6 +379,23 @@ get_numa_node:
 
 	log(TO_CONSOLE, LOG_INFO, "Adding IRQ %d to database\n", irq);
 	return new;
+}
+
+void remove_one_irq_from_db(int irq)
+{
+	struct irq_info find, *tmp;
+	GList *entry = NULL;
+
+	find.irq = irq;
+	entry = g_list_find_custom(interrupts_db, &find, compare_ints);
+	if (!entry)
+		return;
+
+	tmp = entry->data;
+	interrupts_db = g_list_remove(interrupts_db, tmp);
+	free(tmp);
+	log(TO_CONSOLE, LOG_INFO, "IRQ %d was removed from db.\n", irq);
+	return;
 }
 
 static void parse_user_policy_key(char *buf, int irq, struct user_irq_policy *pol)
@@ -585,7 +607,7 @@ static void add_new_irq(char *path, struct irq_info *hint, GList *proc_interrupt
 	/* Set NULL devpath for the irq has no sysfs entries */
 	get_irq_user_policy(path, irq, &pol);
 	if ((pol.ban == 1) || check_for_irq_ban(irq, proc_interrupts)) { /*FIXME*/
-		add_banned_irq(irq, &banned_irqs);
+		__add_banned_irq(irq, &banned_irqs);
 		new = get_irq_info(irq);
 	} else
 		new = add_one_irq_to_db(path, hint, &pol);
--- a/irqbalance.h
+++ b/irqbalance.h
@@ -106,6 +106,8 @@ extern struct irq_info *get_irq_info(int irq);
 extern void migrate_irq(GList **from, GList **to, struct irq_info *info);
 extern void free_cl_opts(void);
 extern void add_cl_banned_module(char *modname);
+extern void add_banned_irq(int irq);
+extern void remove_one_irq_from_db(int irq);
 #define irq_numa_node(irq) ((irq)->numa_node)
 
 
--- a/activate.c
+++ b/activate.c
@@ -48,6 +48,7 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 {
 	char buf[PATH_MAX];
 	FILE *file;
+	int ret = 0;
 
 	/*
  	 * only activate mappings for irqs that have moved
@@ -70,7 +71,12 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 		return;
 
 	cpumask_scnprintf(buf, PATH_MAX, info->assigned_obj->mask);
-	fprintf(file, "%s", buf);
+	ret = fprintf(file, "%s", buf);
+	if (ret < 0) {
+		log(TO_ALL, LOG_WARNING, "cannot change irq %i's affinity, add it to banned list", info->irq);
+		add_banned_irq(info->irq);
+		remove_one_irq_from_db(info->irq);
+	}
 	fclose(file);
 	info->moved = 0; /*migration is done*/
 }


classify.c
```diff
@@ -280,6 +280,13 @@ static void add_banned_irq(int irq, GList **list)
 	return;
 }
 
+#ifdef AARCH64
+void add_banned_list_irq(int irq)
+{
+	add_banned_irq(irq, &banned_irqs);
+}
+#endif
+
```
nhorman: Nothing seems to call this function

procinterrupts.c
```diff
@@ -143,6 +143,36 @@ static void guess_arm_irq_hints(char *name, struct irq_info *info)
...
+
+	fprintf(file, "%s", line);
+	ret = fflush(file);
+
+out:
+	free(line);
+	fclose(file);
+
+	return ret;
+}
```
nhorman: Same here, you've defined two functions, but nothing calls them

procinterrupts.c
```diff
@@ -143,6 +143,36 @@ static void guess_arm_irq_hints(char *name, struct irq_info *info)
...
+	snprintf(buf, PATH_MAX - 1, "/proc/irq/%i/smp_affinity", irq);
+	file = fopen(buf, "r+");
+	if (!file)
+		return -1;
+
+	if (getline(&line, &size, file) <= 0) {
+		ret = -1;
+		goto out;
+	}
+
```
nhorman: How does this even work?  I don't see how being unable to read the smp_affinity of an irq results in deciding that its unchangeable.  If the smp_affinity is unchangeable, it will either sink the write operation and succeed, but not change the contents, or the file will have no writable permissions.  Either way, a failure to read shouldn't have any bearing on its ability to have its affinity set

classify.c
nhorman: This still doesn't make any sense to me.  Why are you not just calling add_banned_irq directly?

procinterrupts.c
```diff
@@ -143,6 +143,36 @@ static void guess_arm_irq_hints(char *name, struct irq_info *info)
...
+	line[strlen(line) - 1] = '\0';
+
+	fprintf(file, "%s", line);
+	ret = fflush(file);
+
```
nhorman: And what is this for?  Why are you writing the same affinity back to the file that you just read it from?  I could sort of understand if you tried to write this value and decided that that a failure in the write operation implied an inability to set the affinity, but all smp_affinity files are read/write for root, so what you would really need to do is look and see if a given irqs smp_affinity changes after you write it.

I'm sorry, I have to ask, are you testing this?

procinterrupts.c
```diff
@@ -143,6 +143,36 @@ static void guess_arm_irq_hints(char *name, struct irq_info *info)
...
+	FILE *file = NULL;
+	char *line = NULL;
+	size_t size = 0;
+	int ret = 0;
+
+	snprintf(buf, PATH_MAX - 1, "/proc/irq/%i/smp_affinity", irq);
+	file = fopen(buf, "r+");
+	if (!file)
+		return -1;
+
```
nhorman: Failure to open the file as r+ could mean many things.  It could mean you're running irqbalance as non-root users.  You need to be careful with this.

procinterrupts.c
```diff
@@ -214,6 +244,24 @@ GList* collect_full_irq_list()
...
+			 * (1) the irq with IRQF_PERCPU flag,  per cpu irq (in arm64, like PPI)
+			 * (2) the irq with IRQD_NO_BALANCING flag, some driver request irq can
+			 *     set the flag according to themselves require. for example in arm64,
+			 *     for the passthrough doorbell irq (GICV4), in future.
+			 * (3) the irq with IRQD_AFFINITY_MANAGED flag, some drivers can set
+			 *     specially irq affinity, and prohibit user to modify it.
+			 *
+			 * For these irqs, we can add these to banned irq list.
+			 */
+			add_banned_list_irq(number);
```
nhorman: Why not just use add_banned_irq here?

classify.c
liuchao173: add_banned_irq is static in classify.c, so it cannot be called directly in procinterrupts.c

procinterrupts.c
liuchao173: banned_irqs and add_banned_irq is static in classify.c

classify.c
nhorman: so you wrote a new function that does the same thing?  why not just bifurcate add_banned_irq to __add_banned_irq(irq, list), and add_banned_irq(irq), and then make add_banned_irq non-static.  Theres no need to create another function with a vaguely simmilar name to do the same work

procinterrupts.c
nhorman: see comment above, theres a better way to do this

procinterrupts.c
liuchao173: I'm testing this in arm server, in which each cpu has a irq called mpt3sas0-msix*. For example, cpu1 has a irq called mpt3sas0-msix1 and its number is 13. In /proc/interrupts:
13:          0   17009314          0          0          0          0          0          0 (omit many zeros) ITS-MSI 575143937 Edge      mpt3sas0-msix1
# cat /proc/irq/13/smp_affinity
00000000,00000002
# ll /proc/irq/13/smp_affinity
-rw-r--r-- 1 root root 0 Mar 13 10:04 /proc/irq/13/smp_affinity
it seem like that smp_affinity is writeable, however is_arm_irq_aff_cannot_change(13) return -1 because write smp_affinity  back and fflush fail. when test arch_timer irq, which smp_affinity is ffffffff,ffffffff, the result is the same. When test normal irqs, like ACPI:Ged or ehci_hcd:usb1, it will success and return 0. And i run the test program as root.
If i write it another smp_affinity  and check if irq's smp_affinity changes, when the irq is normal irq, it will success but it will affect irq and system. i think read old smp_affinity -> write new smp_affinity  -> read new smp_affinity -> write old smp_affinity to restore is too redundancy

classify.c
nhorman: this looks better, but this doesn't need to be an AARCH64 only option, it can be done for all arches, and so avoid the need for the ifdef

irqbalance.h
```diff
@@ -101,6 +101,9 @@ extern int get_cpu_count(void);
 extern void rebuild_irq_db(void);
 extern void free_irq_db(void);
 extern void add_cl_banned_irq(int irq);
+#ifdef AARCH64
+extern void add_banned_irq(int irq);
+#endif
```
nhorman: Same here, with the bifurcation you've done, you can just export this without restricting it to only AARCH64

procinterrupts.c
nhorman: So thats some interesting information.  mpt3sas is an HBA driver, and there should be no reason that the affinity of that devices interrupt is immovable.  The fact that it is suggests a problem in the kernel (though that would surprise me).  What errno value is being returned from fflush?

Additionally, you hit on a real problem with this method of checking for immutable affinity - as I noted, the only way to determine if an irq can't be moved, is to write an affinity mask different from the current one and then rewind (or reopen the file), and read it back comparing the read to the expected value.  If you do that where you've done it, you will change the affinity of the irq (assuming its mutable), which you don't want to do, so this likely isn't the place to make this check.  A better approach would be to do this passively in activate.c.  You could, in activate_mapping, after you've closed the file, reopen it, read the value back, and compare it to info->assigned_obj->mask.  If they differ, then you've found a condition in which an altered affinity for an irq hasn't changed, and you can mark it as such, by logging the error and optionally adding it to the banned list

liuchao173: I don't know how hba works. But i see other irqs that have the same question, such as arm-pmu, they are setted IRQF_NOBALANCING and IRQF_PERCPU， maybe this is the reason

classify.c
nhorman: this looks much better

procinterrupts.c
nhorman: this looks much better

procinterrupts.c
nhorman: fixed

irqbalance.h
nhorman: fixed

classify.c
```diff
@@ -376,6 +381,21 @@ static struct irq_info *add_one_irq_to_db(const char *devpath, struct irq_info *
...
+{
+        struct irq_info find;
+        GList *entry = NULL;
+
+        find.irq = irq;
+        entry = g_list_find_custom(interrupts_db, &find, compare_ints);
+        if (!entry)
+                return;
+
+        interrupts_db = g_list_remove(interrupts_db, entry->data);
```
nhorman: you forgot to free the irq_info struct here (as pointed to by entry->data).  See free_irq_db for a reference to how this is done

liuchao173: yeah = =i have fixed it

Issue #160, opened 2020-08-25
Allow env variable to ban cpus using the cpulist syntax


ChanderG:
I have added code here (https://github.com/ChanderG/irqbalance/commit/677b686e223c74e64f5b230bdb4079555a5a62a9) to basically allow the cpulist syntax ("23", "40-49") to ban cpus from an environment variable.

The existing env variable IRQBALANCE_BANNED_CPUS with the hexadecimal mask is a bit cumbersome to use.

If there is interest in this feature upstream, I can open a PR.

nhorman:
yeah, sure, seems reasonable.  Please open a PR

PR #162, opened 2020-08-26
add env variable to ban cpus using cpulist syntax

ChanderG:
allows banning of cpus using the taskset cpulist syntax in addition to the
bitmap syntax supported via the existing environment variable
IRQBALANCE_BANNED_CPUS


ChanderG:
#160

PR #165, opened 2020-11-19
log correctly for isolated and nohz_full cpus

yeyunfeng-dev:
When the banded CPU is configured, and the "isolcpus" or "nohz_full"
is configured, the log for isolated and nohz CPUs are not correct.

So only log relevant infomation when isolcpus and nohz_full has been
handled, also put the cpus_clear() closer.


ppwaskie (approved):
This seems reasonable to me.

PR #166, opened 2020-11-21
Add IRQBALANCE_BANNED_CPULIST to env file and remove some codes

yeyunfeng-dev:
Add IRQBALANCE_BANNED_CPULIST to env file
Remove IRQBALANCE_BANNED_CPUS environment variable
Remove some unused constant macros in constants.h

### DIFF
--- a/constants.h
+++ b/constants.h
@@ -6,28 +6,4 @@
 
 #define NSEC_PER_SEC 1e9
 
-/* NUMA topology refresh intervals, in units of SLEEP_INTERVAL */
-#define NUMA_REFRESH_INTERVAL 32
-/* NIC interrupt refresh interval, in units of SLEEP_INTERVAL */
-#define NIC_REFRESH_INTERVAL 32
-
-/* minimum number of interrupts since boot for an interrupt to matter */
-#define MIN_IRQ_COUNT	20
-
-
-/* balancing tunings */
-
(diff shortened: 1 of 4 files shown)


nhorman:
removing IRQBALACE_BANNED_CPUs is going to cause breakage for anyone using that environment variable currently on upgrade.  Shouldn't we introduce a deprecation period first?

yeyunfeng-dev:
From the perspective of functional redundancy, it is enough to leave only one environment variable. However, for the consider of compatibility, it is ok to introduce a deprecation period. Of course, if it is removed directly, I think it is also possible for the user to identify the changes,  and handle them with compatibility. 

If the patch "Remove IRQBALANCE_BANNED_CPUS environment variable" not be merged, the other 2 commits is independent patches. Thanks.

nhorman:
yes, I agree.  We could add a deprecation notice as well.  If you will make the requisite changes, I'll merge it all

yeyunfeng-dev:
ok, the patches have been changed

nhorman:
I'm sorry, when I said deprecation notice, I meant for you to write a message into the system log, so that user would be aware of the deprecation, not just individuals reading the code

yeyunfeng-dev:
I have add a log when the environment variable IRQBALANCE_BANNED_CPUS is configured, thanks.

nhorman:
looks good, thanks!

PR #167, opened 2020-12-04
Add irq hotplug feature for irqbalance

hejingxian123:
When new irqs added to the system, irqbalance will need to rescan for rebalancing.
The rescan operation exists the following problems:
1. When there existing many irqs to balance, the rescan operation will cost much time.
2. The rescan operation will produce irq migration with a high probability, which can make irq data cold, and affect the system performance.

To avoid the above problems, we provide the irq hotplug feature for irqbalance. 
When we find new irqs during reading "/proc/interrupts", then we add the new irqs to interrupts_db,
and place them in rebalance irq list. 

The irq hotplug feature for irqbalance can rebalance new irqs quickly without irq migration.We have used the irqbalance hotplug feature in our product for many years.It brings high performance to us when occuring irq hotplug.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -619,7 +619,7 @@ static void add_new_irq(char *path, struct irq_info *hint, GList *proc_interrupt
 /*
  * Figures out which interrupt(s) relate to the device we"re looking at in dirname
  */
-static void build_one_dev_entry(const char *dirname, GList *tmp_irqs)
+static void build_one_dev_entry(const char *dirname, GList *tmp_irqs, int build_irq)
 {
 	struct dirent *entry;
 	DIR *msidir;
@@ -642,10 +642,14 @@ static void build_one_dev_entry(const char *dirname, GList *tmp_irqs)
 			if (!entry)
 				break;
 			irqnum = strtol(entry->d_name, NULL, 10);
-			if (irqnum) {
+			/* If build_irq is valid, only add irq when it's number equals to  build_irq */
+			if (irqnum && ((build_irq < 0) || (irqnum == build_irq))) {
+				printf("add irq:%d %d for %s\n", irqnum, build_irq, path);
 				hint.irq = irqnum;
 				hint.type = IRQ_TYPE_MSIX;
 				add_new_irq(devpath, &hint, tmp_irqs);
+				if (build_irq >= 0)
+					break;
 			}
 		} while (entry != NULL);
 		closedir(msidir);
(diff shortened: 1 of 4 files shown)


classify.c
```diff
@@ -616,6 +616,103 @@ static void add_new_irq(char *path, struct irq_info *hint, GList *proc_interrupt
...
+		if (irqnum == irq)
+			hint.type = IRQ_TYPE_LEGACY;
+	}
+	if (irqnum == irq) {
+		add_new_irq(devpath, &hint, NULL);
+		new = get_irq_info(irq);
+	}
+	return new;
+}
+
```
nhorman, line 660: This is almost an exact duplicate of build_one_dev_entry.  Why are you duplicating so much code?

classify.c
```diff
@@ -616,6 +616,103 @@ static void add_new_irq(char *path, struct irq_info *hint, GList *proc_interrupt
...
+				break;
+			new_irq = build_one_dev_irq(entry->d_name, irq);
+			if (new_irq)
+				break;
+		} while (entry != NULL);
+		closedir(devdir);
+	}
+	return new_irq;
+}
+
```
nhorman, line 681: And this is almost an exact duplicate of rebuild_irq_db.  Instead of cloning functions and suiting them to your needs, find a way to reuse that code

classify.c
```diff
@@ -616,6 +616,103 @@ static void add_new_irq(char *path, struct irq_info *hint, GList *proc_interrupt
...
+{
+	struct irq_info tmp_info = {0};
+	char *savedline;
+
+	savedline = strdup(line);
+	if (!savedline) {
+		return -1;
+	}
+	/* init irq info by parse savedline */
+	init_irq_class_and_type(savedline, &tmp_info, irq);
```
nhorman, line 692: Arent you going to want to use get_irq_user_policy here somewhere?

nhorman:
It looks like a reasonable feature, but you've duplicated lots of code in large sections.  I'll review it more thoroughly once you've managed to re-consolidate those areas, thanks

classify.c
hejingxian123: ok

classify.c
hejingxian123: ok

hejingxian123:
I have refactor build_one_dev_entry function to enable add one irq at one time from the dev dir.
I add new function build_dev_irqs, which can be reused in proc_irq_hotplug and rebuild_irq_db.
At last, I remove the useless banned irq check in proc_irq_hotplug, which has already done in add_new_irq.

hejingxian123:
I have refactored build_one_dev_entry function to enable add one irq at one time from the dev dir.
I add new function build_dev_irqs, which can be reused in proc_irq_hotplug and rebuild_irq_db.
At last, I remove the useless banned irq check in proc_irq_hotplug, which has already done in add_new_irq.

nhorman:
this looks much better thank you, but can you squash the irqs please into one, so we don't have a commit that adds a redundant function, only to have it removed and replaced in the next commit?  That will make it much easier to read.  Thanks!

nhorman:
looks good.  I'll take this as is, but I did notice one other nit.  You added a printf to the code.  That should really be a call to log(), so that irqbalance can redirect those messages to the console or syslog as required by the configuration.  If you could please send another pr correcting that, I would appreciate it.

PR #170, opened 2021-01-04
Add irq hot pull method for irqbalance

hejingxian123:
The pci devices can be hot pulled from the system, and the corresponding irqs will be  deactive. By add the irq hot pull method, we can manage irqs  more efficiently with less resource consumption.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -852,3 +852,40 @@ void sort_irq_list(GList **list)
 {
 	*list = g_list_sort(*list, sort_irqs);
 }
+
+static void remove_no_existing_irq(struct irq_info *info, void *data __attribute__((unused)))
+{
+	GList *entry = NULL;
+
+	if (info->existing) {
+		/* clear existing flag for next detection */
+		info->existing = 0;
+		return;
+	}
+
+	entry = g_list_find_custom(interrupts_db, info, compare_ints);
+	if (entry)
+		interrupts_db = g_list_delete_link(interrupts_db, entry);
+
+	entry = g_list_find_custom(rebalance_irq_list, info, compare_ints);
+	if (entry)
+		rebalance_irq_list = g_list_delete_link(rebalance_irq_list, entry);
+
+	if(info->assigned_obj) {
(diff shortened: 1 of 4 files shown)


nhorman:
I'm sorry, I'm not sure what you're trying to accomplish here.  If you hot remove a pci device, you are correct, the irq will be freed, and on the next update cycle we will rebuild the irq database, with those irqs removed.  What exactly is this change going to accomplish in a faster way?

ppwaskie:
Asked a different way: what is perceived as not working as intended that
needs this change?

hejingxian123:
With the current periodic irq parsing method, irqbalance will never rebuild the irq database when pci devices removed,  Therefore, I think we need the irq hot pull mothod to reduce resource consumption and make active irqs balanced more efficiently .

nhorman:
But why? The entire point of periodic irq parsing  is to detect  changes in our irq list and tebuild.our database.  Can you provide an example of when this won't happen?

hejingxian123:
If need_rescan is false, irqbalance never call free_object_tree and build_object_tree to rebuild the  irq database. Moreover, the parse_proc_interrupts is used to parsing periodically, and the need_rescan flag will not be set true when pci devices removed. 
In the irq parsing function parse_proc_interrupts,  the need_rescan flag will be set true when the following cases  occured:
1、get_irq_info by the irq number readed from /proc/interrupts fail;
2、the new cpu core_count readed from /proc/interrupts is not equal to previous one;
3、the irq_count readed from /proc/interrupts is less than previous, which means the irq has been reinsert.
When the pci devices is removed, the irqs will disappear in /proc/interrupts, and the  irq database cannot be updated by the current method.

nhorman:
ah, that makes more sense, though we should have had a case in proc_parse_interrupts for irqs having gone missing.  I'm sure we did at one point.  I'll have to go through the history and figure out what happened to it.  In either case, This makes more sense to me now, and I'm ok with it.  Thanks for the clarification.

Issue #169, opened 2020-12-24
Interrupt affinity set to a remote NUMA node


shunghsiyu:
On my machine IRQ 4, 8 and 9, which were considered to be on NUMA node 0, were given an smp_affinity of CPUs all on NUMA node 1.

```
$ for i in 4 8 9; do grep . /proc/irq/$i/{node,smp_affinity} /dev/null 2>/dev/null; done
/proc/irq/4/node:0
/proc/irq/4/smp_affinity:ff,ffffc000,000fffff,fc000000
/proc/irq/8/node:0
/proc/irq/8/smp_affinity:ff,ffffc000,000fffff,fc000000
/proc/irq/9/node:0
/proc/irq/9/smp_affinity:ff,ffffc000,000fffff,fc000000
```

This seems strange in two ways:
1. As mentioned, the CPUs in the affinity mask all belong to a remote NUMA node
2. Irqbalance considers them to be on NUMA node -1, which AFAIK means `NUMA_NO_NODE`, and thus should be at least given an affinity mask of all CPUs.
    ```
       Interrupt 9 node_num is -1 (other/1:0)
       Interrupt 8 node_num is -1 (other/1:0)
       Interrupt 4 node_num is -1 (other/1:197)
    ```

Tracing the code I see that `irq_info->numa_node` for IRQ 4/8/9 inside `collect_full_irq_list()` points to NUMA node 0, however that information never made it into `interrupts_db`, probably because `add_new_irq()` is never called on their `irq_info`.

---

The debug log is outputted with e394daf52f42812705761af43322711db04d0513, but I also observe this on v1.7.0 and v1.4.0.

```
$ for i in $(seq 0 300); do grep . /proc/irq/$i/smp_affinity /dev/null 2>/dev/null; done
/proc/irq/0/smp_affinity:ff,ffffffff,ffffffff,ffffffff
/proc/irq/1/smp_affinity:ff,ffffffff,ffffffff,ffffffff
/proc/irq/2/smp_affinity:ff,ffffffff,ffffffff,ffffffff
/proc/irq/3/smp_affinity:ff,ffffffff,ffffffff,ffffffff
/proc/irq/4/smp_affinity:ff,ffffc000,000fffff,fc000000
/proc/irq/5/smp_affinity:ff,ffffffff,ffffffff,ffffffff
/proc/irq/6/smp_affinity:ff,ffffffff,ffffffff,ffffffff
[...]

PR #172, opened 2021-01-28
Also fetch node info for non-PCI devices

ryncsn:
Followring error is observed on some large servers:

[   40.345505] irq 4: Affinity broken due to vector space exhaustion.

Irq 4 is on node 0 but irqbalance put it in node 1. Root cause is that irq 4 is not an from an PCI device so irqbalance failed to get the right numa node number.

This PR fixes this problem.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -361,9 +361,14 @@ static struct irq_info *add_one_irq_to_db(const char *devpath, struct irq_info *
 
 get_numa_node:
 	numa_node = NUMA_NO_NODE;
-	if (devpath != NULL && numa_avail) {
-		sprintf(path, "%s/numa_node", devpath);
-		process_one_line(path, get_int, &numa_node);
+	if (numa_avail) {
+		if (devpath != NULL) {
+			sprintf(path, "%s/numa_node", devpath);
+			process_one_line(path, get_int, &numa_node);
+		} else {
+			sprintf(path, "/proc/irq/%i/node", irq);
+			process_one_line(path, get_int, &numa_node);
+		}
 	}
 
 	if (pol->numa_node_set == 1)


ryncsn:
Hmm, I just noticed there is a issue #169 after I making this PR. That issue is almost the same with the one I'm experiencing.

On my machine after applying this patch the issue is gone. Hopefully I'm getting it right, will this PR will fix the issue?

nhorman:
yeah, I think you did.  We honestly shouldn't have to do this, as the kernel shouldn't issue that error just because the irq is affined to a non-local numa node, but regardless, its better to assign the irq to a local node anyway, so I'll take this.

legoater (commented):
Hi @ryncsn @nhorman , 

While working on a fix related to IPIs of hotplugged CPUs on PPC, I found out that if Linux maps an IRQ associated with an empty node (hence not exposed under `/sys/devices/system/node/`), this change will lead to a core dump in `place_irq_in_node()` because `irq_numa_node(info)` is `NULL`.

It would be good to protect the code with a test on  `place_irq_in_node()` or with something like below : 

```
diff --git a/classify.c b/classify.c
index c08144f432bb..9b4640cd4754 100644
--- a/classify.c
+++ b/classify.c
@@ -376,6 +376,11 @@ get_numa_node:
        else
                new->numa_node = get_numa_node(numa_node);
 
+       if (!new->numa_node) {
+               log(TO_CONSOLE, LOG_WARNING, "IRQ %d has an unknown node\n", irq);
+               new->numa_node = get_numa_node(NUMA_NO_NODE);
+       }
+
        cpus_setall(new->cpumask);
        if (devpath != NULL) {
                sprintf(path, "%s/local_cpus", devpath);
```

What would you prefer  ? 

Thanks,

C.

nhorman:
I think that would be fine, open a PR please

PR #176, opened 2021-04-06
Hotplug may occur again during sleep, so wait until there is no hotplug

liuchao173:


ppwaskie:
Seems reasonable to me.  Thanks for the patch.

PR #178, opened 2021-04-09
Add return value check of opendir in do_one_cpu

hejingxian123:
The input para(path) of do_one_cpu is /sys/devices/system/cpu/cpu$id. When the cpus occur some error, /sys/devices/system/cpu/cpu$id will be diappear, and opendir(path) will return null. Thus, irqbalance will exit with coredump.

### DIFF
--- a/cputree.c
+++ b/cputree.c
@@ -366,7 +366,7 @@ static void do_one_cpu(char *path)
 		struct topo_obj *node;
 
 		dir = opendir(path);
-		do {
+		while (dir) {
 			entry = readdir(dir);
 			if (!entry)
 				break;
@@ -379,8 +379,9 @@ static void do_one_cpu(char *path)
 					break;
 				}
 			}
-		} while (entry);
-		closedir(dir);
+		}
+		if (dir)
+			closedir(dir);
 
 		/*
 		 * In case of multiple NUMA nodes within a CPU package,


cputree.c
```diff
@@ -366,21 +366,23 @@ static void do_one_cpu(char *path)
...
-			entry = readdir(dir);
-			if (!entry)
-				break;
-			if (strncmp(entry->d_name, "node", 4) == 0) {
-				char *end;
-				int num;
-				num = strtol(entry->d_name + 4, &end, 10);
-				if (!*end && num >= 0) {
-					nodeid = num;
+		if (dir) {
```
nhorman, line 369: Instead of nesting the do/while in an if, why not change this the do {}while to :
while (dir && entry) { ...} if(dir) { closedir(dir)}; 
That would save you a level of nesting, which might be nice

hejingxian123:
do {}while cannot change to while (){}, because the entry has not been initialized at the first time.
I have remove the nesting  call.

nhorman:
but....you could have just initialized entry to null.  I.e.

entry = NULL
dir = opendir()
while (dir && entry) {
   ...
}
if (dir)
    closedir()

That way you don't need the extra dir check inside the while loop either

hejingxian123:
if entry=NULL, then condition (dir && entry) will always be false.

nhorman:
Ah, you're right, you can drop the && entry, and make it while (dir) {} .
The if !entry clause in the loop will take care of the rest of it

hejingxian123:
ok

nhorman:
Thank you!

PR #181, opened 2021-04-28
Add basic github workflow based CI

mikey:
This would catch very basic compile issues on x86.

Signed-off-by: Michael Neuling <mikey@neuling.org>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,17 @@
+name: 'test'
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 0 * * 5'
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: ./autogen.sh
+    - run: ./configure
+    - run: make -j$(nproc)


mikey:
This would catch issues like we fixed in #180 before they are merged.

nhorman:
Irabalance already has travis configured.  I think you should just expand that to encompass nonx86 builds.  If you can do that, I'll gladly merge it

mikey:
Arrh, I missed that :facepalm: 

Can you get travis to hooked into github to so we can see the results on each PR on github? I think you just login to travis with the irqbalance github account and click a few things :-)

nhorman:
Its already noted in the README (the build/passing tag).  I can try to put it somewhere more prominent), but I dont think arbitrary markdown is allowed outside of the README file

mikey:
If you look at something like https://github.com/antonblanchard/microwatt/pulls you can see each PR has tick/cross against it. That is what I was thinking. You should be able to get Travis to report there. (The github workflow does that automatically).

nhorman:
grr, its apparently a bit more complicated than clicking a few buttons, especially for irqbalance that is owned as part of an orgnization rather than by me personally.  I'll look into it shortly

neheb:
@mikey IMO should use something like setup-alpine which has a lot more architectures.

nhorman:
you're welcome to propose a pr if you like

PR #183, opened 2021-05-07
remove no existing irq in banned_irqs

liuchao173:
when a banned irq doesn't exist, it won't be removed from banned_irqs


liuchao173:
Yes, because banned irqs are in banned_irqs, not in interrupts_db. And we can't judge where this info is in remove_no_existing_irq

PR #186, opened 2021-07-17
Add spread-mode

TieNan2019:
Sparse all irqs all over according to cpu topology, as tested, sometimes this may gain better performance.

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -34,7 +34,7 @@ struct user_irq_policy {
 	int numa_node;
 };
 
-static GList *interrupts_db = NULL;
+GList *interrupts_db = NULL;
 static GList *banned_irqs = NULL;
 GList *cl_banned_irqs = NULL;
 static GList *cl_banned_modules = NULL;
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -50,6 +50,7 @@ char socket_name[64];
 int one_shot_mode;
 int debug_mode;
 int foreground_mode;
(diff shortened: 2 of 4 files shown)


TieNan2019:
Greetings, maintainer!

The reason why I wrote this is that after a few tests. I find it interesting that sometime just simply sparsing irqs all over, we would gain much better performance, even better compared with that used “irqbalance --oneshot”.

In this test, I used Alibaba baremetal server, and redis-benchmark command to verify this conclusion. It seems that sometimes, like when we just boot up the system or the irqbalance cannot handle the distribution of irq loads well, it would be a better idea to fix irqs on cpus sparsely. I got the idea from this script: https://ecs-image-tools.oss-cn-hangzhou.aliyuncs.com/ecs_mq/ecs_mq_latest.tgz。

Surprising, this script works foolish but works pretty in many situation. But this script doesn't take cpu topology into consideration, I guess that's something worth doing.

Thanks,
Weisson.

nhorman:
I'm sorry,but this patch is just not acceptable.   It creates a policy that is based at best on anecdotal reaults.   If you want to support this policy, please include your benchmark reaults worh and without your policy.  Even if it is  better,  we dont typically implement custom policies in the code.  We have the policyscript option for that.  We can implement this spread policy as a script and include it in the trees contrib directory

nhorman:
closing for lack of activity

PR #188, opened 2021-08-01
configure.ac: use pkg-config to find numa

ffontaine:
Use pkg-config to find numa and fallback to current mechanism. Thanks to pkg-config, numa dependencies such as `-latomic` will be retrieved.

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,7 +27,7 @@ EXTRA_DIST = COPYING autogen.sh misc/irqbalance.service misc/irqbalance.env
 SUBDIRS = tests
 
 UI_DIR = ui
-AM_CFLAGS = $(LIBCAP_NG_CFLAGS) $(GLIB2_CFLAGS)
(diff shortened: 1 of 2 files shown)


nhorman:
I wouldn't bother with the fallback mechanism.  If libnuma has a pkg-config file, let's just use it.

Also there is no reason. To do the AC_SUBST calls, PKG_CHECK_MODULES will handle that (see the glib2 use)

ffontaine:
`numa.pc` is only available since version 2.0.12 and https://github.com/numactl/numactl/commit/53ac79b8d3256a225c18a25f4046798000bec862 so I think it is a good idea to keep a fallback.

I'll remove the `AC_SUBST` calls.

nhorman:
Thank you!  2 years is my general rule for not worrying about a fallback mechanism.  If a system can't update their libraries for two years, they can figure out compatibility.

PR #190, opened 2021-08-23
fix opendir fails in check_platform_device

liuchao173:
When irq name does not contain spaces, savedptr is an empty string and irq_fullname will have a extra space at the end like "LNRO0005:00 ".
So opendir in check_platform_device will fail, and irqbalance prints log:
"No directory /sys/devices/platform/LNRO0005:00 /: No such file or directory"

### DIFF
--- a/procinterrupts.c
+++ b/procinterrupts.c
@@ -178,10 +178,12 @@ void init_irq_class_and_type(char *savedline, struct irq_info *info, int irq)
 	}
 
 #ifdef AARCH64
-	snprintf(irq_fullname, PATH_MAX, "%s %s", last_token, savedptr);
-	tmp = strchr(irq_fullname, '\n');
-	if (tmp)
-		*tmp = 0;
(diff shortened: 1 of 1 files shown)


nhorman:
Can you post what your /proc/interrupts file looks like here?  I'm not sure what you mean when you say irq_name does not contain spaces. init_irq_class_and_type parses a full line of that file at a time, and so there should always be a space available to consume in that parsing

liuchao173:
90:          0          0          0          0  mbigen-v2 839 Edge      HISI0213:1f
91:          0          0          0          0  mbigen-v2 848 Edge      HISI0243:00
92:          0          0          0          0  mbigen-v2 849 Edge      HISI0243:01

When irq_name is Edge or Level, it breaks the loop. And lasttoken is "HISI0213:1f" and savedptr is "". So irq_fullname is "HISI0213:1f "

nhorman:
gotcha, thank you

PR #201, opened 2022-04-04
Confine irqbalance to systems where it is useful.

jcpunk:
Systems with only one CPU cannot benefit from this service, so
don't start it when it isn't worth the resource consumption.


jcpunk:
Interesting, my config management keeps trying to start the service on my 1 core vm (admittedly I should probably do some more confines on my end, but thought sending this upstream would be a helpful thing I should do)....

nhorman:
well, it will try to start the service, but it will fail, since the check is in main.  I guess it doesn't hurt anything to provide the constraint to systemd

PR #206, opened 2022-05-17
Support thermal events

ChangSeokBae:
Hello irqbalance folks,

I previously attempted to share this patchset via the mailing list. But
since there has been no feedback, I've decided to give my pull request
here. The patches essentially support interaction with hardware to
determine which CPUs to be masked from IRQ handling. This approach is
different from user-driven CPU masking that is already in irqbalance.
Here is a little bit of background and some details about that.
Appreciate your feedback.

== Thermal Events ==

Intel's Hardware Feedback Interface (HFI) provides a monitoring mechanism
for a CPU's performance and energy efficiency status. A table is shared
between hardware and the operating system. The table is updated as per
operating condition changes, e.g. due to thermal limit.

The table has numeric values to indicate the CPU's capabilities. When some
critical condition is exhibited, userspace is expected to cooperate for
controlling resources. If both performance and efficiency values are
dropped to zero, the CPU is significantly lagging in handling interrupts.
Thus, any further interrupt handling should be avoided from it.

== Event Subscription ==

The kernel [1] provides a Netlink interface for userspace to subscribe to
the HFI table update. The series provides to subscribe and handle the
events for properly re-distributing interrupts in a hardware-interactive
way. The Netlink is a socket-base connection. The connection with the
'irqbalance-ui' process is also established in a socket.

== CPU Banning Approach ==

The CPU banning takes place on the daemon's polling. So, hysteresis in
reacting to thermal events is predictable. Preemptive interrupt
re-distribution may reduce the reaction time. But it involves changes in
the main loop. Also, the benefit will be marginal as long as the compute
demand is not so fluctuating in that thermal-constrained levels.

== Patchset ==

The patchset begins with a couple of preparatory patches to introduce
helpers and establish build support with required libraries. Then,
the implementation follows for each Netlink-specific helper to establish
an initial handshake. Finally, an HFI-specific event handler is in place
for masking CPUs properly.

[1] HFI support in the kernel:
https://lore.kernel.org/lkml/20220127193454.12814-1-ricardo.neri-calderon@linux.intel.com/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -36,6 +36,28 @@ AS_IF([test "x$has_ncursesw" = "xyes"], [
   AC_SUBST([LIBS])
 ])
 
+AC_CANONICAL_HOST
+
+AC_ARG_ENABLE(thermal,
+  AS_HELP_STRING([--enable-thermal], [enable thermal event support [default=auto]]),,
+  AS_IF([test x"$host_cpu" = x"x86_64"], [enable_thermal=yes], [enable_thermal=no])
+)
+
+AS_IF([test x"$enable_thermal" = x"yes" && test x"$host_cpu" != x"x86_64"],
+  AC_MSG_ERROR([no thermal events support on $host_cpu systems.]),
+)
+
+AS_IF([test x"$enable_thermal" = x"yes"],
+  [PKG_CHECK_MODULES([LIBNL3], [libnl-3.0 libnl-genl-3.0], [have_thermal=yes],
+    AC_MSG_NOTICE([no thermal event support as libnl-3.0 is unavailable.])
+  )]
+)
+
+AS_IF([test "x$have_thermal" = xyes],
+  AC_DEFINE([HAVE_THERMAL], 1, [Build irqbalance to support thermal events])
+)
+AM_CONDITIONAL([THERMAL], [test "x$have_thermal" = xyes])
+
 AC_C_CONST
 AC_C_INLINE
 AM_PROG_CC_C_O
--- a/thermal.c
+++ b/thermal.c
@@ -1,0 +1,83 @@
+/*
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 as
+ * published by the Free Software Foundation.
+ */
+
+#include <stdio.h>
+
+#include <netlink/genl/genl.h>
+#include <netlink/genl/family.h>
+#include <netlink/genl/ctrl.h>
+
+#include "irqbalance.h"
+
+cpumask_t thermal_banned_cpus;
+
+static gboolean prepare_netlink(void)
+{
+	gboolean error = TRUE;
+
+	log(TO_ALL, LOG_ERR, "thermal: not yet implement to alloc memory for netlink.\n");
+	return error;
+}
+
+#define NL_FAMILY_NAME	"nlctrl"
+
+static gboolean establish_netlink(void)
+{
+	gboolean error = TRUE;
+
+	log(TO_ALL, LOG_ERR, "thermal: not yet implemented to establish netlink.\n");
+	return error;
+}
+
+static gboolean register_netlink_handler(nl_recvmsg_msg_cb_t handler __attribute__((unused)))
+{
+	gboolean error = TRUE;
+
+	log(TO_ALL, LOG_ERR, "thermal: not yet implemented to register thermal handler.\n");
+	return error;
+}
+
+static gboolean set_netlink_nonblocking(void)
+{
+	gboolean error = TRUE;
+
+	log(TO_ALL, LOG_ERR, "thermal: not yet implemented to set nonblocking socket.\n");
+	return error;
+}
+
+void deinit_thermal(void)
+{
+	return;
+}
+
+/*
+ * return value: TRUE with an error; otherwise, FALSE
+ */
+gboolean init_thermal(void)
+{
+	gboolean error;
+
+	error = prepare_netlink();
+	if (error)
+		goto err_out;
+
+	error = establish_netlink();
+	if (error)
+		goto err_out;
+
+	error = register_netlink_handler(NULL);
+	if (error)
+		goto err_out;
+
+	error = set_netlink_nonblocking();
+	if (error)
+		goto err_out;
+
+	return FALSE;
+err_out:
+	deinit_thermal();
+	return TRUE;
+}
--- a/thermal.c
+++ b/thermal.c
@@ -14,43 +14,342 @@
 
 cpumask_t thermal_banned_cpus;
 
+#define INVALID_NL_FD	-1
+#define MAX_RECV_ERRS	2
+
+#define THERMAL_GENL_FAMILY_NAME	"thermal"
+#define THERMAL_GENL_EVENT_GROUP_NAME	"event"
+#define NL_FAMILY_NAME			"nlctrl"
+
+struct family_data {
+	const char *group;
+	int id;
+};
+
+static struct nl_sock *sock;
+static struct nl_cb *callback;
+
+/*
+ * return value: TRUE with an error; otherwise, FALSE
+ */
 static gboolean prepare_netlink(void)
 {
-	gboolean error = TRUE;
+	int rc;
 
-	log(TO_ALL, LOG_ERR, "thermal: not yet implement to alloc memory for netlink.\n");
-	return error;
+	sock = nl_socket_alloc();
+	if (!sock) {
+		log(TO_ALL, LOG_ERR, "thermal: socket allocation failed.\n");
+		return TRUE;
+	}
+
+	rc = genl_connect(sock);
+	if (rc) {
+		log(TO_ALL, LOG_ERR, "thermal: socket bind failed.\n");
+		return TRUE;
+	}
+
+	callback = nl_cb_alloc(NL_CB_DEFAULT);
+	if (!callback) {
+		log(TO_ALL, LOG_ERR, "thermal: callback allocation failed.\n");
+		return TRUE;
+	}
+
+	return FALSE;
 }
 
-#define NL_FAMILY_NAME	"nlctrl"
+static int handle_groupid(struct nl_msg *msg, void *arg)
+{
+	struct nlattr *attrhdr, *mcgrp, *cur_mcgrp;
+	struct nlattr *attrs[CTRL_ATTR_MAX + 1];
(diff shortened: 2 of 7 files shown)


thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	if (!sock) {
+		log(TO_ALL, LOG_ERR, "thermal: socket allocation failed.\n");
+		goto err_out;
+	}
 
-	log(TO_ALL, LOG_ERR, "thermal: not yet implement to alloc memory for netlink.\n");
+	rc = genl_connect(sock);
+	if (rc) {
+		log(TO_ALL, LOG_ERR, "thermal: socket bind failed.\n");
+		goto err_out;
```
nhorman, line 95: memory leak, need to free the allocated socket

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	}
+
+	callback = nl_cb_alloc(NL_CB_DEFAULT);
+	if (!callback) {
+		log(TO_ALL, LOG_ERR, "thermal: callback allocation failed.\n");
+		goto err_out;
```
nhorman, line 101: again, free the socket in the error case

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	}
+
+	goto out;
+err_out:
```
nhorman, line 105: Just set error=false here and fall through rather than having an extra goto

nhorman:
sorry, that mailing list has been closed for years, I'm surprised you didn't get a bounce.  All development happens here now.

Generally This patch looks ok to me, but you're going to need to alter the conditions under which thermal event handling is included.  From the configure.ac file, it appears you enable the thermal handling code if libnl3 is available, but that can be true on many arches on which irqbalance runs (like arm/mips.etc), which won't have thermal support.  While that doesn't look catastrophic from a build time perspective, it will result in errors printed when the genl socket family can't be looked up, and that's bad.  I would suggest you gate the inclusion of thermal support on the target CPU being x86_64 and libnl3 presence.  You may also want to add an opt-out switch to allow for building on smaller embedded x86 platforms that don't want thermal support.

I've made some extra comments in the code, but generally, it looks like a useful feature

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+		goto err_out;
+	}
+
+	cloned_callback = nl_cb_clone(callback);
+	if (!cloned_callback) {
+		log(TO_ALL, LOG_ERR, "thermal: callback handle duplication failed.\n");
+		goto err_out;
+	}
+
+	rc = nl_send_auto_complete(sock, msg);
```
nhorman, line 253: FYI, not a huge deal but nl_send_auto_complete is deprecated in libnl3, probably want to use nl_send_auto instead

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	}
 
-	log(TO_ALL, LOG_ERR, "thermal: not yet implemented to establish netlink.\n");
+	group_id = nldata.id;
+	if (group_id < 0) {
+		log(TO_ALL, LOG_ERR, "thermal: invalid group_id was received.\n");
+		goto err_out;
+	}
+
+	rc = nl_socket_add_membership(sock, group_id);
```
nhorman, line 297: If any of these operations fails, what code is responsible for tearing down the socket?

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	rc = nl_cb_set(callback, NL_CB_VALID, NL_CB_CUSTOM, handle_thermal_event, NULL);
+	if (rc) {
+		error = TRUE;
+		log(TO_ALL, LOG_ERR, "thermal: thermal handler registration failed.\n");
+	}
+out:
 	return error;
 }
 
+#define INVALID_NL_FD	-1
```
nhorman, line 453: what is this doing here?

thermal.c
ChangSeokBae: These are in init_thermal() where it falls through deinit_thermal() that frees up sock.

thermal.c
ChangSeokBae: Given this, I thought the code is more readable than doing like this down there:

	fd = nl_socket_get_fd(sock);
	if (fd == -1) { ...

thermal.c
```diff
@@ -14,43 +14,333 @@
...
+	gboolean received = FALSE;
+
+	if (condition == G_IO_IN) {
+		int rc;
+
+		rc = nl_recvmsgs(sock, callback);
+		if (!rc)
+			received = TRUE;
+		else
+			log(TO_ALL, LOG_ERR, "thermal: failed to receive thermal event.\n");
```
nhorman, line 468: IIRC g_unit_fd_add cancels the subscribed watch if you return FALSE here.  Which is fine, but it means that a failed call to nl_recvmsgs (which will happen if any of your handler fail to parse the message), will cause the entire monitoring system to shut down.  Is that what you want?

ChangSeokBae:
> I would suggest you gate the inclusion of thermal support on the target CPU
> being x86_64 and libnl3 presence. You may also want to add an opt-out
> switch to allow for building on smaller embedded x86 platforms that don't
> want thermal support.

Ah, surely, looks to be all valid points. Thanks.

thermal.c
nhorman: I'm fine with that, its just in the middle of the code.  Please move it to the top of the file or in front of the function in which you use it

ChangSeokBae: Oh, I see. Will do that.

thermal.c
ChangSeokBae: Yeah, good point. Let me think a little bit more on this.

ChangSeokBae: So, if it returns FALSE here, it will remove the source. Even if a message carries NLMSG_ERROR, or its handling fails, the subscription itself should be separately established. There are some side effects of abandoning it but it is okay to keep it I think.

So let me change not to avoid returning FALSE with some comment like this.

ChangSeokBae:
Sorry for my delay on this though, I just updated my repository to address your comments.

 Hopefully that fulfills your request. But let me know if I'm missing anything here.

thermal.c
ChangSeokBae: In the end, I made it returns FALSE after a few failures. I came to consider a possibility to disconnect it. (See the updated code.)

configure.ac
```diff
@@ -36,6 +36,31 @@ AS_IF([test "x$has_ncursesw" = "xyes"], [
...
+    AS_IF([test x"$has_libnl3" = x"yes"], [
+      AC_MSG_NOTICE([support thermal event.])
+      support_thermal=yes;],
+      AC_MSG_NOTICE([no thermal event support without libnl-3.0.])),
+    AC_MSG_NOTICE([thermal event support is disabled with --enable-thermal=no])
+  )
+)
+AS_IF([test "x$support_thermal" = xyes], [
+  AC_DEFINE([HAVE_THERMAL], 1, [Build irqbalance to support thermal events])
+])
```
nhorman, line 61: This seems a bit backwards in terms of logic:

- PKG_CHECK_MODULES([LIBNL3]) only needs to be done if we select support thermal, so it should be moved inside the AS_IF([... support_thermal]) condition
- The 3 nested AS_IF conditions can just be a single test inside AS_IF([support thermal])
- You don't need both an enable_thermal and a support_thermal variable.
The whole thing can be simplified to the following pseudo code:

AC_ARG_ENABLE[--enable_thermal]
AS_IF([test <enable_thermal = yes> -a <host_cpu != x8664], [
     AC_MSG_ERROR([we don't support thermal events on non-intel platforms])], [])
AS_IF([test <enable_thermal = yes>], [
   PKG_CHECK_MODULES([LIBNL3]....)])

nhorman: Sorry, you also need to add the AC_DEFINE([HAVE_THERMAL]) in the action-if-found clause of PKG_CHECK_MODULES

thermal.c
nhorman: ?

thermal.c
nhorman: this is still deprecated

thermal.c
ChangSeokBae: There is a fall-though path: prepare_netlink()->deinit_thermal()->nl_socket_free(sock)

thermal.c
ChangSeokBae: In this fall-through path: prepare_netlink()->deinit_thermal()->nl_cb_put(callback), it frees up.

thermal.c
ChangSeokBae: Sorry, I missed this. I don't recall why..

Yeah, I now think these whole gotos are not necessary. Simply return TRUE on every failure. And the fall-through returns FALSE. Then explain the implication of this boolean value at the function comment.

Will set the resolve button along with my revision.

thermal.c
ChangSeokBae: Will do. Sorry for missing this too.

configure.ac
ChangSeokBae: It sounds like you want the default of that option 'no', right?  Otherwise, it will always return error on non-x86 64-bit systems.

nhorman: No, I'm fine with thermal events being available by default, but you can do that with the AC_ARG_ENABLE macro - i.e. use the action-if-not-given clause to set thermal_enable=$enableval iff test host_cpu == x86_64

The rest is just to consolidate the validity checking and make it a bit more readable

ChangSeokBae: Okay, then the default is variable.  If user wants to set this on non-x86 systems, this error shows up; otherwise, will be smooth as the default is switched to no here.

I guess it should work. Let me check this under Qemu's ARM emulator.

thermal.c
ChangSeokBae: Updated: https://github.com/Irqbalance/irqbalance/compare/3ec98d963fa3a483073e73a73998c2f317dabfc5..560291389e33db108d2fb6d954ef059f953c6e33#diff-3066472ef74ce9f800762e2831b21aea1cd51c554c719742f0ece2f74c2e4a03L256

configure.ac
ChangSeokBae: Okay, I found it works. You can see the changes from this line in the diff: https://github.com/Irqbalance/irqbalance/compare/3ec98d963fa3a483073e73a73998c2f317dabfc5..560291389e33db108d2fb6d954ef059f953c6e33#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R43

Issue #202, opened 2022-04-05
--banmod not respected


johalun:
Hi

I'm running Ubuntu 20.04.3 with hwe-5.11 kernel and irqbalance 1.6.0. I want to disable irq balancing for our Chelsio network cards so I run it with `--banmod cxgb4` option combined with pinning one IRQ / CPU for the two Chelsio NICs. However, IRQs are still registered across all CPUs for both NICs (seen in /proc/interrupts). If I disable irqbalance completely I get the desired behavior of one IRQ per CPU.

Am I using the option incorrectly or what might be the issue?

nhorman:
I don't know.  can you run irabalance manually with the foreground and debug options enabled along with the --banmod option you are setting, then upload the debug output here?  Also if you could upload /proc/interrupts that would be helpful

johalun:
The driver cxgb4 from ChelsioUwire package, not the one bundled with the kernel. Not yet tested with inbox driver.

```
   sudo irqbalance --debug --banmod cxgb4
Isolated CPUs: 00000000
Adaptive-ticks CPUs: 00000000
Banned CPUs: 00000000
Package 0:  numa_node 0 cpu mask is 00ffffff (load 0)
        Cache domain 0:  numa_node is 0 cpu mask is 00200200  (load 0)
                CPU number 21  numa_node is 0 (load 0)
                CPU number 9  numa_node is 0 (load 0)
        Cache domain 1:  numa_node is 0 cpu mask is 00800800  (load 0)
                CPU number 11  numa_node is 0 (load 0)
                CPU number 23  numa_node is 0 (load 0)
        Cache domain 2:  numa_node is 0 cpu mask is 00080080  (load 0)
                CPU number 7  numa_node is 0 (load 0)
                CPU number 19  numa_node is 0 (load 0)
        Cache domain 3:  numa_node is 0 cpu mask is 00040040  (load 0)
                CPU number 18  numa_node is 0 (load 0)
                CPU number 6  numa_node is 0 (load 0)
        Cache domain 4:  numa_node is 0 cpu mask is 00020020  (load 0)
                CPU number 5  numa_node is 0 (load 0)
                CPU number 17  numa_node is 0 (load 0)
        Cache domain 5:  numa_node is 0 cpu mask is 00010010  (load 0)
                CPU number 16  numa_node is 0 (load 0)
                CPU number 4  numa_node is 0 (load 0)
[...]

PR #208, opened 2022-06-07
get irq->module relationship from /sys/bus/pci/*/driver

liuchao173:
https://github.com/Irqbalance/irqbalance/issues/202

Signed-off-by: Chao Liu <liuchao173@huawei.com>

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -7,6 +7,7 @@
 #include <dirent.h>
 #include <assert.h>
 #include <errno.h>
+#include <libgen.h>
 
 #include "irqbalance.h"
 #include "types.h"
@@ -578,7 +579,7 @@ static int check_for_module_ban(char *name)
 		return 0;
 }
 
-static int check_for_irq_ban(int irq, GList *proc_interrupts)
+static int check_for_irq_ban(int irq, char *mod, GList *proc_interrupts)
 {
 	struct irq_info find, *res;
 	GList *entry;
@@ -594,6 +595,9 @@ static int check_for_irq_ban(int irq, GList *proc_interrupts)
 	/*
 	 * Check to see if we banned module which the irq belongs to.
 	 */
+	if (mod != NULL && strlen(mod) > 0 && check_for_module_ban(mod))
+		return 1;
+
 	entry = g_list_find_custom(proc_interrupts, &find, compare_ints);
 	if (entry) {
 		res = entry->data;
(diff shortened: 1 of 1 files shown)


classify.c
```diff
@@ -604,19 +607,42 @@ static int check_for_irq_ban(int irq, GList *proc_interrupts)
...
 
+static void get_mod_name(char *path, char *mod)
+{
+	char *cmd;
+	FILE *output;
+
+	if (!path)
+		return;
+
+	cmd = alloca(strlen(path) + 30);
```
nhorman, line 618: why +30 here?

classify.c
```diff
@@ -594,6 +594,9 @@ static int check_for_irq_ban(int irq, GList *proc_interrupts)
 	/*
 	 * Check to see if we banned module which the irq belongs to.
 	 */
+	if (strlen(mod) > 0 && check_for_module_ban(mod))
```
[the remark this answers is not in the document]
liuchao173: mod is defined in add_new_irq as char array，it won't be null

classify.c
```diff
@@ -604,19 +607,42 @@ static int check_for_irq_ban(int irq, GList *proc_interrupts)
...
+	if (!cmd)
+		return;
+
+	sprintf(cmd, "cat %s/uevent | grep \"DRIVER=\"", path);
+	output = popen(cmd, "r");
+	if (!output)
+		return;
+
+	fgets(mod, 128, output);
+	pclose(output);
```
nhorman, line 628: why go to all the trouble of fishing the DRIVER field out of the event file?  you can find the same information via basename(readlink(/sys/bus/pci/*/driver)

classify.c
liuchao173: It is prepared for the remaining string "cat %s/uevent | grep \"DRIVER=\""

classify.c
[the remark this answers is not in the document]
nhorman: thats even worse, by defining it on the stack without memsetting it to anything, you're passing around whatever garbage is on the stack when the call is made, so you avoid the semi-guaranteed segfault, in exchange for a random segfault that occurs when strlen indexes through whatever random memory you have that likely doesn't have a NULL terminator.

classify.c
nhorman: yeah, but 30?  Thats just a random sizing of the string.  at least define your template and take its strlen, i.e.:
#define CMDTMPL "/uevent | grep \"DRIVER=\""
...
cmd = alloca(strlen(path) + strlen(CMDTMPL) + 1); //alloc cmd template for path + null terminator byte

classify.c
liuchao173: done

PR #211, opened 2022-06-12
add meson

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

### DIFF
--- a/contrib/README
+++ b/contrib/README
@@ -1,0 +1,2 @@
+This directory contains meson build instructions for irqbalance. This is here to see if there is any interest from
+the general community.
--- a/contrib/meson.build
+++ b/contrib/meson.build
@@ -1,0 +1,45 @@
+project('irqbalance', 'c', version: '1.9.0', default_options: ['warning_level=1'])
+cc = meson.get_compiler('c')
+
+glib_dep = dependency('glib-2.0')
+m_dep = cc.find_library('m', required: false)
+capng_dep = dependency('libcap-ng', required: get_option('capng'))
+ncurses_dep = dependency('curses', required: get_option('ui'))
+systemd_dep = dependency('libsystemd', required: get_option('systemd'))
+
+cdata = configuration_data()
+cdata.set('HAVE_GETOPT_LONG', cc.has_function('getopt_long'))
+cdata.set('HAVE_IRQBALANCEUI', ncurses_dep.found())
+cdata.set('HAVE_NUMA_H', cc.has_header('numa.h'))
+cdata.set('HAVE_LIBCAP_NG', capng_dep.found())
+cdata.set('HAVE_LIBSYSTEMD', systemd_dep.found())
+cdata.set_quoted('VERSION', meson.project_version())
+cfile = configure_file(output: 'config.h', configuration: cdata)
+
+if cdata.get('HAVE_IRQBALANCEUI')
+  add_project_arguments('-D_GNU_SOURCE', language: 'c')
+
+  executable(
+    'irqbalance-ui',
+    '../ui/helpers.c',
+    '../ui/irqbalance-ui.c',
+    '../ui/ui.c',
+    dependencies: [glib_dep, ncurses_dep],
+    install: true,
+  )
+endif
+
+executable(
+  'irqbalance',
+  '../activate.c',
+  '../bitmap.c',
+  '../classify.c',
+  '../cputree.c',
+  '../irqbalance.c',
+  '../irqlist.c',
+  '../numa.c',
+  '../placement.c',
+  '../procinterrupts.c',
+  dependencies: [glib_dep, m_dep, capng_dep, systemd_dep],
+  install: true,
+)
--- a/contrib/meson_options.txt
+++ b/contrib/meson_options.txt
@@ -1,0 +1,11 @@
+option('capng', type : 'feature',
+  description : 'Build with libcap-ng support',
+)
+
+option('systemd', type : 'feature',
+  description : 'Build with systemd support',
+)
+
+option('ui', type : 'feature',
+  description : 'Build the UI component',
+)


nhorman:
Why?  We already have a mature and functional build system.  I really don't don't understand the need for another to maintain and update just because someone likes it

neheb:
My motivation was to get rid of various hacks regarding using a static glib2, which can use an external iconv and gettext. When so, linker flags must be added manually.

https://github.com/openwrt/packages/pull/18733

nhorman:
1) what exactly are the static linking issues with with glib2?  Your commit makes no attempt to describe what problem you are trying to solve regarding any linking issues.  Your commit just says "add meson" which is irritatingly vague.  If you really wanted to fix some build issue, I would have thought you would have at least laid out the problem in the commit message.

2) again, why meson.  What problem with static linking is so egregious that you can't solve it with autoconf/automake?

neheb:
well, the issue is that I personally don't speak autotools so I don't know what the proper way to fix it there is.

The flags that need to be added: https://github.com/openwrt/packages/blob/master/utils/irqbalance/Makefile#L53

It is so because it's not handling static glib properly.

nhorman:
ok, I get that, but the solution is not to add another build system to solve what otherwise seems a pretty easy problem

looking at your openwrt link, it _seems_ like openwrt is trying to do build of irqbalance but hass added their own option to just link glib2 statically?  is that the goal here?  I ask because I can add an option to build irqbalance statically, but _just_ linking glib2 in statically is a hack in my view that im not comfortable with

neheb:
Yeah for size reasons this is done. glib2 is huge and LTO can get rid of a decent amount of the size.

In meson it's basically

```
dependency('glib')
```

vs

```
dependency('glib', static: true)
```

I assume it's more work with autotools.

nhorman:
Not especially more difficult (about 5-10 lines in the makefile.am), but I am philosophically opposed to offering per library granularity in the use of static vs dynamic libs during link.  I get the desire for a static application, but it's either a static app or it's not, I'm not trying to support /maintain an application that has a random collection of static and dynamic libs.

neheb:
right. well, this PR doesn't do any of those shenanigans. It just introduces meson, which is smaller and more readable than the autotools one. It also has better defaults (FILE_OFFSET_BITS=64, fPIC for shared libraries, ccache used by default when available, etc...)

nhorman:
You're right it.doesn't, I don't see any fixes for any linking, or even any changes to how linking is done differently between the autotools environment and the meson environment

Which begs the question, what are you actually fixing here? 

I really don't appreciate having someone post PRs that make significant changes to projects I maintain, with no explanation, and then some general hand waving when I ask about it.

What is your real goal here?  Are you just interested in building this project with meson instead of autotools?  We can have that debate, but I'm going to need.alot of convincing since autotools works perfectly well for me, as well as numerous others who are currently building it for various distributions and platforms.  Changing the build environment will have to provide significant advantages without causing significant hardship for existing users.  Note I'm saying this under the auspices that I'll maintain a single build system.  Supporting two is a recepie for disaster and added unneeded work

If on the other hand, you're really trying to fix a linker error, that's something we can address, but at this point I'm going to need.you to describe the problem in depth so we can understand and fix it properly within the existing build system.  I'm sorry, but I'm not going to support/replace a new build system for some linker issue you've only vaguely described

neheb:
```
Are you just interested in building this project with meson instead of autotools?
```

doing so would be best. In OpenWrt, a lot of autotools projects need to have parallel compilation disabled as not all of them build as such. All meson projects in OpenWrt build in parallel as ninja has no such problems. The same was done with cmake there once cmake was switched to compile with ninja instead of Makefiles.

```
I don't see any fixes for any linking, or even any changes to how linking is done differently between the autotools environment and the meson environment
```

The goal is to see nothing as meson already handles this perfectly. No hacks are needed.


```
Changing the build environment will have to provide significant advantages without causing significant hardship for existing users.
```

Significant advantages from OpenWrt's standpoint are:
Faster compilation: No autoreconf is needed, guaranteed safe parallel compilation, proper linking if glib was built with external libiconv.

```
If on the other hand, you're really trying to fix a linker error, that's something we can address, but at this point I'm going to need.you to describe the problem in depth so we can understand and fix it properly within the existing build system.
```

meson mostly deals with pkgconfig files so I'll explain it like this. glib-2.0.pc:

```
Name: GLib
Description: C Utility Library
Version: 2.70.5
Requires.private: libpcre >=  8.31
Libs: -L${libdir} -lglib-2.0 -lintl
Libs.private: -pthread -liconv -lm
Cflags: -I${includedir}/glib-2.0 -I${libdir}/glib-2.0/include
```

That's with external libiconv and gettext. Without, it looks like:

```
Name: GLib
Description: C Utility Library
Version: 2.70.5
Requires.private: libpcre >=  8.31
Libs: -L${libdir} -lglib-2.0
Libs.private: -pthread -lm
Cflags: -I${includedir}/glib-2.0 -I${libdir}/glib-2.0/include
```

As is, irqbalance's autotools build does not add -liconv when necessary, this is worked around in the OpenWrt Makefile. That workaround unfortunately will not work soon as external iconv will be going away soon. Instead of doing a dance around this, it's better to handle it properly.

nhorman:
This is just so much more hand waving

> ```
> Are you just interested in building this project with meson instead of autotools?  

> ```
> 
> doing so would be best. In OpenWrt, a lot of autotools projects need to have parallel compilation disabled as not all of them build as such. All meson projects in OpenWrt build in parallel as ninja has no such problems. The same was done with cmake there once cmake was switched to compile with ninja instead of Makefiles.

This is non-sensical.  autotools works perfectly fine with parallel make.  I just ran make -j in irqbalance on my 40-way system and it built just fine and very quickly.  To justify replacing the build system of a specific project because "lots of" other unnamed projects have some vague undefined problem with parallel make suggests to me that you haven't really thought about what you're trying to fix or improve here.

> 
> ```
> I don't see any fixes for any linking, or even any changes to how linking is done differently between the autotools environment and the meson environment
> ```
> 
> The goal is to see nothing as meson already handles this perfectly. No hacks are needed.
> 
You haven't even identified the problem that you're really trying to solve, so I don't see how you can assert what changes are needed, or even if they could be considered hacks or appropriate fixes.

> ```
> Changing the build environment will have to provide significant advantages without causing significant hardship for existing users.
> ```
> 
> Significant advantages from OpenWrt's standpoint are: Faster compilation: No autoreconf is needed, guaranteed safe parallel compilation, proper linking if glib was built with external libiconv.
> 

- Faster is a hand waving argument.  To put some real numbers to it, ninja will build irqbalance in .214 seconds of real time, while autotools does it in .401 seconds.  Strictly speaking, yes faster, and for a larger project that might be significant, but you're talking about .1 seconds of savings in exchange for having a significant number of users having to re-integrate a new build environment into their systems.  Not enough of a benefit in my mind
- Lack of need for autoreconf is an invalid argument.  meson requires a setup step, just like autotools does.
- Guaranteed safe parallel compilation is invalid - you haven't identified what the problem is with parallel make in autotools, and it works fine for me.  If you want to address this we can, but just aserting it is not something I'm going to accept.
- proper linking - you're still hand waving, identify in detail what the problem is, or stop complaining about it.



> ```
> If on the other hand, you're really trying to fix a linker error, that's something we can address, but at this point I'm going to need.you to describe the problem in depth so we can understand and fix it properly within the existing build system.
> ```
> 
> meson mostly deals with pkgconfig files so I'll explain it like this. glib-2.0.pc:
> 
> ```
> Name: GLib
> Description: C Utility Library
> Version: 2.70.5
> Requires.private: libpcre >=  8.31
> Libs: -L${libdir} -lglib-2.0 -lintl
> Libs.private: -pthread -liconv -lm
> Cflags: -I${includedir}/glib-2.0 -I${libdir}/glib-2.0/include
> ```
> 
> That's with external libiconv and gettext. Without, it looks like:
> 
> ```
> Name: GLib
> Description: C Utility Library
> Version: 2.70.5
> Requires.private: libpcre >=  8.31
> Libs: -L${libdir} -lglib-2.0
> Libs.private: -pthread -lm
> Cflags: -I${includedir}/glib-2.0 -I${libdir}/glib-2.0/include
> ```
> 
> As is, irqbalance's autotools build does not add -liconv when necessary, this is worked around in the OpenWrt Makefile. That workaround unfortunately will not work soon as external iconv will be going away soon. Instead of doing a dance around this, it's better to handle it properly.

Ok, so we're actually getting somewhere now.  Libs.private (according to man file.pc), indicates that Libs.private only specifies libraries needed during static linking (which IIRC is what you want).  But irqbalance doesn't currently support static linking, so those libraries never get included (which is correct for a dynamic link)

so all you really want here, is the ability to link irqbalance statically without having to introduce openwrt specific makefile rules.  Thats doable.  I've created a PR here for you to test with:
https://github.com/Irqbalance/irqbalance/pull/212
It adds an --enable-staticbuild option to the configure script, which, when selected sets up PKG_CHECK_MODULES to behave like PKG_CHECK_MODULES_STATIC (implying the reporting of Libs.private in various pkgconfig files, and directs the makefile to do a static link.  I've not tested it mind you because I don't have the static version of all the libraries installed, but if you have them installed for openwrt, this should work.

neheb:
I shouldn't write the next section but I'll indulge:

> Faster is a hand waving argument. To put some real numbers to it, ninja will build irqbalance in .214 seconds of real time, while autotools does it in .401 seconds. Strictly speaking, yes faster, and for a larger project that might be significant, but you're talking about .1 seconds of savings in exchange for having a significant number of users having to re-integrate a new build environment into their systems. Not enough of a benefit in my mind

well then:

autotools:

```
time make package/irqbalance/compile
Executed in   19.78 secs    fish           external
   usr time   16.56 secs  276.00 micros   16.56 secs
   sys time    2.62 secs   52.00 micros    2.62 secs
```

meson:

```
time make package/irqbalance/compile
Executed in   13.03 secs    fish           external
   usr time   11.32 secs    0.00 micros   11.32 secs
   sys time    2.06 secs  398.00 micros    2.06 secs
```

It's not just the raw compilation time. There's also autoreconf.

nhorman:
The contents of autogen.sh (like meson setup) are meant to.be a once per.platfo execution (ie once run for a given environment, you should never need to run it again, ditto configure once unless your build configuration changes).  The fact that your environment reruns it everytime.ypu want to compile is a cause to make your build.ebvirment more efficient, not for me to overhaul my build environment.
 
Also, what are you building on?  I can run autogen.sh; ./configure; make -j in 8.9 seconds of wall clock time.  I think you said your environment (I assume openwrt) is serializing the build. I still don't see any reason why you would want/need.to do that.  You've alluded.to so me parallel make problem, but I don't see what it is, and you could cut your itqbalance  build time in half by just making that change without any replacement of autotools

neheb:
It's full on OpenWrt buildroot. Those compilation time are misleading as the build system checks that the dependencies are built and if not, build them (They were built in my test). But at the end of the day, it's still a difference of 6.75 seconds, mostly because of autoreconf. I ran the build serially on purpose. I tried as best as I could to compare A and B.

The parallel issues have to do with some but not all autotools packages. I haven't seen such issues with irqbalance.

Anyway, I guess the only thing left here is to close this PR...

nhorman:
I'm sorry, I'm kind of lost for words here.  [Here](https://github.com/Irqbalance/irqbalance/pull/211#issuecomment-1153918360) we determined that there was about 0.1 seconds difference in compilation time.  [Here](https://github.com/Irqbalance/irqbalance/pull/211#issuecomment-1155967240) you're claiming a much bigger difference based on needing to rerun one-time-only operations on every build.  Thats a limitation of openwrts build environment, not a problem to throw over the wall to me.

As this point, I think you just want meson, because its easier for you.  You've opened this PR without any explanation until I pulled it out of yout, then hand waved some excuses about parallel builds and static linking, and have finally settled on speed as your reasoning, as the most cogent argument you could muster.  I don't appreciate being asked to make changes just because its easier for one person to work with.  Especially when said change means I need to maintain two independent build systems moving forward.

That said, I'm not a Luddite, and I'm not opposed to change, But I need evidence that its going to make life easier for a large group of people beyond just you.  So I'll make this proposal to you: Modify this PR such that the meson build files are placed in a contrib directory, along with a readme explaining that its an currently unsupported build environment, but is being added here as an experiment to see if there is general community uptake/interest. We can watch distributions for the next 12 months (giving them time to adopt a later version between minor releases), and if we see that distribution packaging converts to the use of meson over autotools, and that community contributions make a point to modify the meson build files, we can consider the community to have interest in meson here, and work on an autotools deprecation strategy, replacing it with meson.

Make those changes, and I'll merge this

neheb:
Done.

thesamesam:
FWIW, I've adopted it for the Gentoo packaging as of 1.9.4 (I didn't realise it was there until now, as I only usually look in the top-level dir for build systems). Arch seems to have adopted it as well.

nhorman:
Apologies, I'm away at the moment, will review this shortly

nhorman:
Ok, you appear to be correct.  If you want to submit a PR to move the meson build file to the root directory, I'm ok with it.

eli-schwartz:
Done: #211

Issue #218, opened 2022-06-28
irqbalance-ui: F5 tab cpulist column overrides class column


vishal14051992:
Hello,

In order to better explain this issue, I am attaching screenshot.

https://i.imgur.com/d23i5C7.png

Here for example IRQ0,  cpulist can be used instead of individual cpu
like 0-14,32-47 or 'ASSIGNED TO CPUS' can be kept as last column.

It will also be helpful if device/interrupt name can also be displayed.
This will help in easily mapping irqN with name of device.

Thanks,

vishal14051992:
Hello Neil,

I tested recent code and I can see that cpulist is moved to last column. It will be very user friendly if the tab only show cpulist range. For example, instead of below output for IRQ0,

NUMBER          IS BANNED        CLASS                           ASSIGNED TO CPUS
 
   IRQ 0           NO               Other                           0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 32, 33, 34, 35, 36, 37, 38, 39, 40, 41, 42, 43, 44, 45, 46, 47

it should show,

NUMBER          IS BANNED        CLASS                           ASSIGNED TO CPUS
 
   IRQ 0           NO               Other                                         0-15,32-47

This will save a lot of space and will provide more crisp output for large systems with 200+ cpus.

Thanks,

PR #219, opened 2022-06-28
irqbalance-ui: move 'ASSIGNED TO CPUS' to the last column

liuchao173:
https://github.com/Irqbalance/irqbalance/issues/218


vishal14051992:
Hello Neil,

Thank you for quickly sharing the patch.
I also wanted to ask if the same tab can include names of interrupt/device just like we see in last column of /proc/interrupts file.
This will help to easily map irqN with device name.

liuchao173:
irqbalance-ui obtains interrupt information through socket. Currently, names of interrupt/device are not included. I will try to add it if Neil agrees. @nhorman

nhorman:
I'm fine with adding that, though given the amount of information we're presenting in the UI, it might also be time to look at refactoring the ncurses table into a different format to be more accommodating of long cells

PR #221, opened 2022-06-30
fix memory leak in ui/ui.c

liuchao173:


liuchao173:
I am confused too.

nhorman:
Ack, looks much better, merge when ready

PR #223, opened 2022-07-11
irqbalance-ui: print cpulist in SETUP IRQS

liuchao173:
save space for printing interrupt names
https://github.com/Irqbalance/irqbalance/issues/224

### DIFF
--- a/ui/ui.c
+++ b/ui/ui.c
@@ -352,9 +352,32 @@ void handle_cpu_banning()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


liuchao173:
In the old version, only the first 64 CPUs can be printed due to the size of assigned_to. Is it necessary to change 128 to NR_CPUS 4096?

liuchao173:
@nhorman

PR #226, opened 2022-07-14
irqbalance-ui: display irq name in SETUP IRQS

liuchao173:
get irq name from /proc/interrupts

### DIFF
--- a/ui/ui.c
+++ b/ui/ui.c
@@ -9,6 +9,8 @@ int max_offset;
 GList *all_cpus = NULL;
 GList *all_irqs = NULL;
 
+static char **irq_name;
+
 char *IRQ_CLASS_TO_STR[] = {
 			"Other",
 			"Legacy",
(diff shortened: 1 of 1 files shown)


liuchao173:
I think passing interrupt names through sockets increases length and changes a lot. Running the script once per interrupt may seem silly. It should be OK to run the script separately after printing interrupts and then print the names of all interrupts.

liuchao173:
@nhorman Please review.

liuchao173:
> looks good, 2 review comments left unaddressed however

I didn't see the review comment.

nhorman:
https://github.com/Irqbalance/irqbalance/pull/226/commits/f771d7193f8562a5bdbc4dfec3b8cdf25cdea9df

They're listed above as well

liuchao173:
“0 comments on commit f771d71”
I can't find them

nhorman:
sigh, stupid ideosyncracy of github apparently.  From this window, click on the commits tab, then click on the one commit in the PR.  They're also listed as pending in the review in the conversation we're having above

liuchao173:
I did exactly what you said, but I still didn't see it = =

nhorman:
I don't know what to tell you, I did a review, and I see it in the commits area clear as day

nhorman:
jeez.

line 415 of ui.c:
use fscanf here instead of fgets and sscanf. That way you can error check your read to make sure you didn't overrun your 128 byte buffer.

line 420 of ui.c:
This, like the other call that was removed is potentially prone to overruns. use a double call to snprintf here with the first too args set to NULL,0 to get the formatted string size which you can then calloc/alloca

liuchao173:
done

Issue #228, opened 2022-07-15
irqbalance: debug shows "IRQ0 was BANNED" if --banirq= is used without any value


vishal14051992:
# ./irqbalance --banirq= -d -o
IRQ 0 was BANNED.      <=----
This machine seems not NUMA capable.
Prevent irq assignment to these isolated CPUs: 00000000
Prevent irq assignment to these adaptive-ticks CPUs: 00000000
Banned CPUs: 00000000
....
....
Adding IRQ 138 to database
Adding IRQ 139 to database
IRQ 0 was BANNED.     <=----
Adding IRQ 4 to database
Adding IRQ 8 to database

No impact as such, but I do not observe this issue with irqbalance-1.0.7-12.el7.x86_64.
Nonetheless, manually writing to irq0 affinity is not allowed, but 'IRQ 0 was BANNED' 
seem to be misleading.

PR #230, opened 2022-07-18
irqbalance: check the length of optarg

liuchao173:
https://github.com/Irqbalance/irqbalance/issues/228

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -124,6 +124,7 @@ static void parse_command_line(int argc, char **argv)
 	int opt;
 	int longind;
 	unsigned long val;
+	char *endptr;
 
 	while ((opt = getopt_long(argc, argv,
 		"odfjVi:p:s:c:l:m:t:e:",
@@ -139,8 +140,8 @@ static void parse_command_line(int argc, char **argv)
 				exit(1);
 				break;
 			case 'c':
-				deepest_cache = strtoul(optarg, NULL, 10);
-				if (deepest_cache == ULONG_MAX || deepest_cache < 1) {
+				deepest_cache = strtoul(optarg, &endptr, 10);
+				if (optarg == endptr || deepest_cache == ULONG_MAX || deepest_cache < 1) {
 					usage();
 					exit(1);
 				}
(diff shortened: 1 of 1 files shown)


nhorman:
what exactly are you fixing here?  I understand wanting to validate your input prior to using it, but getopt_long already does that.  Note the lopts struct and option_string that is passed to getopt long.  For every case in the switch statement, the corresponding option character and lopt entry is followed by a ':' and a 1 in the has_arg field respectively, indicating the option has a required argument.  In the event a valid argument is not provided getopt_long returns '?', leading us to print usage and exit.  I don't see what this patch is improving

liuchao173:
For example, when use "--banirq=" without any value, optarg is an empty string and val will be 0. It will cause IRQ 0 is banned.
see the issue:
https://github.com/Irqbalance/irqbalance/issues/228

nhorman:
ok, my point remains.  Please look at the function call for getopt_long:
  while ((opt = getopt_long(argc, argv,
                "odfjVi:p:s:c:l:m:t:e:",
                lopts, &longind)) != -1) {

the option string contains the substring "i:", and the banirq line in lopts is:
{"banirq", 1 , NULL, 'i'},
indicating a required argument

if getopt_long is returning 'i' when you provide --banirq= on the command line, getopt_long is the problem.  getopt_long should either be returning '?" in that case, or (in the case its returning 'i') should set optarg to "" (i.e. a non-null string that has no characters in it, a pointer to '\0').  In either case checking for NULL shouldn't be the solution here

liuchao173:
--banirq=, --banirq="", -i ""
"" is also a argument. getopt_long returning 'i' and set optarg to "". I use strlen to check it rather than checking for NULL.
How about checking the value of val and don't ban IRQ0

nhorman:
We're saying the same thing.  If you pass "" as the argent.to -i, and
getopt_long returns 'i', then optarg should be set to a non null pointer to
a string who's first element is '\0'.

That said, using strlen shouldn't be needed there (though it likely
works).  You can probably still use strtoull with the second arg set to
non-null, using that to determine an error in parsing

liuchao173:
I get it

PR #240, opened 2022-10-18
Avoid uninitialized read

dirkmueller:
irq_info struct is initialized on stack so the members need to be initalized to avoid a crash on uninitialized pointer dereference.

Signed-off-by: Dirk Müller <dirk@dmllr.de>

### DIFF
--- a/classify.c
+++ b/classify.c
@@ -259,7 +259,7 @@ static gint compare_ints(gconstpointer a, gconstpointer b)
 
 static void __add_banned_irq(int irq, GList **list)
 {
-	struct irq_info find, *new;
+	struct irq_info find = {0}, *new;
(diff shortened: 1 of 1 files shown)


dirkmueller:
this can be easily found by e.g using `-fanalyzer`:

```
classify.c: In function ‘check_for_irq_ban’:
classify.c:602:16: warning: use of uninitialized value ‘*irq.name’ [CWE-457] [-Wanalyzer-use-of-uninitialized-value]
  602 |         if (irq->name != NULL && strlen(irq->name) > 0 && check_for_module_ban(irq->name))
      |             ~~~^~~~~~
  ‘rebuild_irq_db’: events 1-2

```

Issue #242, opened 2022-10-25
thermald integration on AMD cpus leads to an error entry in journal


dirkmueller:
The journal entry for irqbalance on an AMD host is a bit too verbose: 

```

Oct 24 08:35:16 Mobile-PC /usr/sbin/irqbalance[3546]: thermal: socket bind failed.
```

this is raised as an *error* level, while on an AMD x86_64 system there is no thermald. it would be nice if it would detect that situation and not log it as error or not log it at all.

dirkmueller:
commit was merged, thank you!

PR #244, opened 2022-10-26
change  the log level in thermal.c from error to warning

liuchao173:
https://github.com/Irqbalance/irqbalance/issues/242

### DIFF
--- a/thermal.c
+++ b/thermal.c
@@ -99,13 +99,13 @@ static gboolean prepare_netlink(void)
 
 	rc = genl_connect(sock);
 	if (rc) {
-		log(TO_ALL, LOG_ERR, "thermal: socket bind failed.\n");
+		log(TO_ALL, LOG_INFO, "thermal: socket bind failed, thermald may not be running.\n");
 		return TRUE;
 	}
 
 	callback = nl_cb_alloc(NL_CB_DEFAULT);
 	if (!callback) {
-		log(TO_ALL, LOG_ERR, "thermal: callback allocation failed.\n");
+		log(TO_ALL, LOG_WARNING, "thermal: callback allocation failed.\n");
 		return TRUE;
 	}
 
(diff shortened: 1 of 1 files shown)


thermal.c
```diff
@@ -93,19 +93,19 @@ static gboolean prepare_netlink(void)
 
 	sock = nl_socket_alloc();
 	if (!sock) {
-		log(TO_ALL, LOG_ERR, "thermal: socket allocation failed.\n");
+		log(TO_ALL, LOG_WARNING, "thermal: socket allocation failed.\n");
```
dirkmueller, line 96: I think these are valid errors, as they should not fail unless extreme circumstances

thermal.c
```diff
@@ -93,19 +93,19 @@ static gboolean prepare_netlink(void)
...
 		return TRUE;
 	}
 
 	rc = genl_connect(sock);
 	if (rc) {
-		log(TO_ALL, LOG_ERR, "thermal: socket bind failed.\n");
+		log(TO_ALL, LOG_WARNING, "thermal: socket bind failed.\n");
```
dirkmueller, line 102: this one is the one that you get when there is no thermald available. imho even LOG_INFO with slightly more useful hint (something like `thermal: socket bind failed, is thermald running?`) could be useful here imho.

thermal.c
```diff
@@ -372,7 +372,7 @@ static int handle_thermal_event(struct nl_msg *msg, void *arg __attribute__((unu
 		THERMAL_GENL_ATTR_MAX,	/* maximum attribute id as expected */
 		NULL);			/* validation policy */
 	if (rc) {
-		log(TO_ALL, LOG_ERR, "thermal: failed to parse message for thermal event.\n");
+		log(TO_ALL, LOG_WARNING, "thermal: failed to parse message for thermal event.\n");
```
dirkmueller, line 375: same here, I think those are valid errors

PR #257, opened 2023-02-02
Fix CPU number condition in service file

middlingphys:
Modified to start only when there are 2 or more CPUs, as intended in the original pull request (https://github.com/Irqbalance/irqbalance/pull/201).


ms178:
@nhorman This change causes a warning in dmesg with the following line:

`[  +0,028967] systemd[1]: /usr/lib/systemd/system/irqbalance.service:6: Missing '=', ignoring line.`

Seen with systemd-git 253.r62916.df40b7ef5a

middlingphys:
I was misunderstanding the syntax of the unit file. The original line was reflecting the condition as intended. I will submit a pull request to revert the change.

PR #262, opened 2023-03-09
Issue 259: select NL_SKIP / NL_STOP based on error

nhorman:
the handle_error function for thermal should skip EINTR errors, but stop for everything else


ppwaskie (approved):
LGTM

PR #264, opened 2023-04-19
procinterrupts: fix initialisation of regex_t struct

nekopsykose:
```
{NULL} utilises the null pointer, but this is not valid, because null is a pointer:

procinterrupts.c:110:29: error: initialization of 'long unsigned int' from 'void *' makes integer from pointer without a cast [-Werror=int-conversion]
  110 |                 { "eth.*" ,{NULL} ,NULL, IRQ_TYPE_LEGACY, IRQ_GBETH },

0-initialisation should be done with '0' instead of a pointer.
```


nekopsykose:
thanks!

PR #265, opened 2023-07-04
activate_mapping: fflush the buffered data to smp_affinity

liutgnu:
Previously irqbalance uses the return value of fprintf() to decide whether the modification of smp_affinity is successful or not. However it is not reliable because fprintf() is stream buffered, a fflush() should be used to check the buffered data been successfully written into the file before the judgement.

This patch fixes the issue by introducing fflush() after fprintf().

### DIFF
--- a/activate.c
+++ b/activate.c
@@ -76,7 +76,7 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 
 	cpumask_scnprintf(buf, PATH_MAX, applied_mask);
 	ret = fprintf(file, "%s", buf);
-	if (ret < 0) {
+	if (ret < 0 || fflush(file)) {
 		log(TO_ALL, LOG_WARNING, "cannot change irq %i's affinity, add it to banned list", info->irq);
 		add_banned_irq(info->irq);
 		remove_one_irq_from_db(info->irq);


activate.c
```diff
@@ -76,7 +76,7 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 
 	cpumask_scnprintf(buf, PATH_MAX, applied_mask);
 	ret = fprintf(file, "%s", buf);
-	if (ret < 0) {
+	if (ret < 0 || fflush(file)) {
```
rjarry, line 79: Hi there,

would it be possible to log the value of `strerror(errno)` in the warning message as well?

rjarry:
Hi Neil,

this is intended to address the issue I reported here: https://bugzilla.redhat.com/show_bug.cgi?id=2184735

TL;DR:

```
~# echo 0,20 > /proc/irq/47/smp_affinity_list
-bash: echo: write error: No space left on device
```

nhorman:
ok, that makes sense,thanks

rjarry:
Oh, I thought you would wait for @liutgnu to respond to my comment. Also, I had another remark.

activate.c
rjarry: Also, it may be better to check for `fflush(file) < 0`.

liutgnu: Hi Robin, thanks for your comments! Since Neil has merged the patch, for 1) I think we can get it improved later if users really care about the failing reason of IRQs reassigning. for 2) according to man 3 fflush, when successful 0 is returned, non-zero is other cases. So zero checking and negative checking are same to me. Thanks!

rjarry: I think it is crucial to know what is the **exact** reason why the IRQ affinity cannot be enforced.

About `fflush(3)`, this is not exactly what I have in my man page (fedora 38):

```
RETURN VALUE
       Upon successful completion 0 is returned.
       Otherwise, EOF is returned and errno is set to indicate the error.
```

By the way, I wonder if there could also be an error returned by `fclose` which is currently ignored:

https://github.com/Irqbalance/irqbalance/blob/v1.9.2/activate.c#L84

Would that be OK if I send another PR with these two minor changes?

rjarry: Here it is #266

Issue #267, opened 2023-07-07
Coredumps (Kernel 6.3.12)


Uglymotha:
On my system ruinning 6.3.12 kernel, irqbalance coredumps with a SIGSEGV

[  832.859285] traps: irqbalance[467787] general protection fault ip:55b99ffd5e9d sp:7ffe76006820 error:0 in irqbalance[55b99ffcb000+d000]

poll([{fd=4, events=POLLIN}, {fd=5, events=POLLIN}], 2, 9940) = 0 (Timeout)
openat(AT_FDCWD, "/proc/interrupts", O_RDONLY) = 7
newfstatat(7, "", {st_mode=S_IFREG|0444, st_size=0, ...}, AT_EMPTY_PATH) = 0
read(7, "            CPU0       CPU1     "..., 1024) = 1024
read(7, "0          0          0         "..., 1024) = 1024
read(7, "    0          0          0     "..., 1024) = 1024
read(7, "    0          0          0  PCI"..., 1024) = 1024
read(7, "         0          0          0"..., 1024) = 1024
read(7, "        0  IR-PCI-MSI-0000:00:03"..., 1024) = 1024
read(7, "       0          0          0  "..., 1024) = 1024
read(7, "-PCI-MSI-0000:20:01.1    0-edge "..., 1024) = 1024
read(7, "     0          0          0    "..., 1024) = 1024
read(7, ":20:07.1    0-edge      PCIe PME"..., 1024) = 1024
read(7, "   0          0          0      "..., 1024) = 1024
read(7, "-edge      PCIe PME\n  56:       "..., 1024) = 1024
read(7, "        0          0          0 "..., 1024) = 1024
read(7, "    0-edge      nvme2q0\n  67:   "..., 1024) = 1024
read(7, "     0          0          0    "..., 1024) = 1024
read(7, "2\n  73:          0          0   "..., 1024) = 1024
read(7, "          0          0          "..., 1024) = 1024
read(7, "0          0          0         "..., 1024) = 1024
read(7, "          0          0          "..., 1024) = 1024
read(7, "         0          0          0"..., 1024) = 1024
read(7, "        0          0          0 "..., 1024) = 1024
read(7, "      0          0          0   "..., 1024) = 1024
read(7, "     0          0          0    "..., 1024) = 1024
read(7, "    0          0          0     "..., 1024) = 1024
read(7, "   0          0          0      "..., 1024) = 1024
read(7, "  0          0          0       "..., 1024) = 1024
[...]

PR #268, opened 2023-07-10
Revert "arm64: Add irq aff change check"

rjarry:
This reverts commit 55c5c321c73e4c9b54e041ba8c7d542598685bae.

The banned_irqs list is meant to be configured by the user, it is not a dynamic list. Also, remove_one_irq_from_db() was freeing an irq_info struct which was still referenced in banned_irqs and potentially in rebalance_irq_list.

Iterating on any of these lists will cause a use after free error.

Fixes: https://github.com/Irqbalance/irqbalance/issues/267

### DIFF
--- a/activate.c
+++ b/activate.c
@@ -48,7 +48,6 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 {
 	char buf[PATH_MAX];
 	FILE *file;
-	int ret = 0;
 	cpumask_t applied_mask;
 
 	/*
@@ -75,12 +74,7 @@ static void activate_mapping(struct irq_info *info, void *data __attribute__((un
 		return;
 
 	cpumask_scnprintf(buf, PATH_MAX, applied_mask);
-	ret = fprintf(file, "%s", buf);
-	if (ret < 0) {
-		log(TO_ALL, LOG_WARNING, "cannot change irq %i's affinity, add it to banned list", info->irq);
-		add_banned_irq(info->irq);
-		remove_one_irq_from_db(info->irq);
-	}
+	fprintf(file, "%s", buf);
 	fclose(file);
 	info->moved = 0; /*migration is done*/
 }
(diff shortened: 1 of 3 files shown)


nhorman:
This was already done after other reports were opened.about a segfault

rjarry:
Hi Neil,

I saw that you reverted another commit which revealed this use after free. But The original bug remains. If `fprintf()` returns `< 0` here: https://github.com/Irqbalance/irqbalance/blob/master/activate.c#L78 You will have the same issue.

This patch is a revert of the root cause: https://github.com/Irqbalance/irqbalance/commit/55c5c321c73e4c9b54e041ba8c7d542598685bae

Can you reconsider and apply this revert?

Thanks.

nhorman:
that said, just reverting this commit isn't sufficient.  This commit fixes a real problem (namely that aarch64 has interrupts whos affinity can't be moved, despite their affinity files being writeable.  There needs to be some early detection of these irqs so that they can be safely added to the banned irq list (or otherwise ignored)

rjarry:
I completely agree. However the original commit does **not** fix that "aarch64 has interrupts whos affinity can't be moved" issue. Otherwise you would have hit that use after free earlier.

I can have a look at fixing the aarch64 problem in a different way.

nhorman:
That doesn't make it right to just revert it and punt the real fix to a later time.  In the state the the tree is in now, it works properly (minus the most recent issue you found).  The path forward is not to revert the old commit, it is to find a new change that fixes your reported behavior without breaking the intent of the old fix

rjarry:
Ok, I'll close this and open another PR with a proper fix then.

nhorman:
either way, you're welcome to update this one if you prefer.  FWIW, I'd look at adding a generic test in rebuild_irq_db to identify interrupts that fail writes to their corresponding smp_affinity file, and just not add them, or mark them as otherwise unavailable.  This would also serve as a method of identifying any other interrupts that fail  affinity setting for other reasons (such as the one you've identified)

PR #277, opened 2023-11-09
service: invoke with --journal

Flowdalic:
irqbalance has a --journal option to make efficient use of systemd's journaling API. Let's use that.


nhorman:
I'm not opposed to this change, but I'm curious to know what exactly it provides.  If you're using a systemd unit file, systemd will redirect all stdout/stderr output to the journal anyway.  This option was really meant for when you're running irqbalance manually, and wanted output to go to the journal.

nhorman:
closing, seems to be a duplicate of #278

PR #278, opened 2023-11-09
Automatically use systemd journal logging if run under systemd

Flowdalic:
This is an alternative to https://github.com/Irqbalance/irqbalance/pull/277

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -617,6 +617,16 @@ int main(int argc, char** argv)
 	parse_command_line(argc, argv);
 
 	/*
+	 * Check if we are run under systemd and enable journal_logging
+	 * and run in foreground if so. Systemd v232 or later will set
+	 * INVOCATION_ID.
+	 */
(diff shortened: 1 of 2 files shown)


nhorman:
as with your other PR, What benefit does this provide?  if we're running as a systemd service, all stdout/stderr output gets sent to the journal anyway.

Flowdalic:
> if we're running as a systemd service, all stdout/stderr output gets sent to the journal anyway.

It's the other way around: if you run as a systemd service, then why would you use stdout/stderr if you could directly systemd *vectorized* (and thread-safe, but I guess that is irrelevant for irqbalance) API to submit log messages?

Granted, it is a small win, but a win nevertheless.

BTW, a new irqbalance release would be nice. The latest release contains a suboptimal service file (missing AF_NETLINK).

nhorman:
Given that irqbalance runs once every 10 seconds or so, I don't see that as much of a win, but sure, it doesn't hurt anything either

PR #280, opened 2023-11-13
Thermal support and NUMA support as feature when building with meson

dvzrv:
This is a quick follow-up to #279, where I realized only afterwards, that the NUMA support should likely also be a feature instead of always on and that the thermal support via libnl had not been added yet.

For the thermal support `thermal.c` is only added to sources, when the dependency is actually detected.


contrib/meson.build
```diff
@@ -5,13 +5,13 @@ glib_dep = dependency('glib-2.0')
 m_dep = cc.find_library('m', required: false)
 capng_dep = dependency('libcap-ng', required: get_option('capng'))
 ncurses_dep = dependency('curses', required: get_option('ui'))
-numa_dep = dependency('numa')
+numa_dep = dependency('numa', required: get_option('numa'))
```
nhorman, line 8: So, i've obviously tipped my hand previously.  I'm not near as familiar with meson as I am with autotools.  These look like a pretty straightforward mapping of autotools options to meson build options, but in autotools --with-numa is the default (ditto --with-irqbalance-ui).  Shouldn't these be defaulted to on, with an option to disable?  Or is the default value set somewhere that I'm not seeing?

contrib/meson.build
```diff
@@ -5,13 +5,13 @@ glib_dep = dependency('glib-2.0')
...
 systemd_dep = dependency('libsystemd', required: get_option('systemd'))
 
 cdata = configuration_data()
 cdata.set('HAVE_GETOPT_LONG', cc.has_function('getopt_long'))
 cdata.set('HAVE_IRQBALANCEUI', ncurses_dep.found())
-cdata.set('HAVE_NUMA_H', cc.has_header('numa.h'))
+cdata.set('HAVE_NUMA_H', cc.has_header('numa.h') and numa_dep.found())
 cdata.set('HAVE_LIBCAP_NG', capng_dep.found())
 cdata.set('HAVE_LIBSYSTEMD', systemd_dep.found())
```
nhorman, line 19: I understand you didn't write it this way, but autotools default enables systemd support if the library is located, and offers an option to disable it.  If we're starting to consider a switch to meson, shouldn't there be an option to disable it as well?

contrib/meson.build
dvzrv: >  I'm not near as familiar with meson as I am with autotools.

Well, I'm not judging :)

>  Shouldn't these be defaulted to on, with an option to disable? Or is the default value set somewhere that I'm not seeing?

These are configured in meson_options.txt, representing the [build options](https://mesonbuild.com/Build-options.html#build-options) for the project. In our particular case here we have [features](https://mesonbuild.com/Build-options.html#features), which can be `auto`, `enabled` or `disabled` (when building you would e.g. pass `-Dthermal=enabled` to `meson` and downstream on Arch Linux we in fact make use of `--auto-features enabled`, which switches all features to `enabled`).

Currently, the `meson_options.txt` does not specify any default and I think this means to default to `auto` which is apparently the same as saying: `required: false`, when the `required: get_option('numa')` is resolved.

contrib/meson.build
dvzrv: The build option for systemd integration is exposed via the entry in `meson_options.txt`.
As mentioned above, one would either specify the preference via `meson -Dsystemd=enabled`, or by using `--auto-features enabled` (to enable all features).

Currently, the workflow is structured in such a way, that all wanted features need to be toggled on (unless we would add a `value: enabled` to the entries in `meson_options.txt`). This is because the `get_option()` call in the [dependency()](https://mesonbuild.com/Reference-manual_functions.html#dependency) call further up in this file defines whether the systemd dependency is searched for.

contrib/meson.build
nhorman: Can it be changed so that numa is defaulted to enabled and we can disable it on the command line?

dvzrv: Yes. This can be added in the meson_options.txt.
I'll adapt it tomorrow! :)

dvzrv: Sorry, took me much longer than anticipated, as I fell ill last week.

I have now adapted the `meson_options.txt` (should be renamed to `meson.options` in the future) to enable the NUMA feature by default.

contrib/meson.build
dvzrv: I have enabled the systemd feature by default in an additional commit.

nhorman:
LGTM

Issue #292, opened 2024-01-05
Log spam in journald


cpaelzer:
Hi,
Ubuntu got this reported as a bug in [2046470](https://bugs.launchpad.net/ubuntu/+source/irqbalance/+bug/2046470).
A check of the history showed me that this happened

1. [back in the initial import this was a debug message](https://github.com/Irqbalance/irqbalance/commit/95f9881ff8d1032ae903bf700c17d6855bb1dacf#diff-74a30482d921ab3bffa86b0b1d5d481f70bff90c7ace5089813c521b98be07efR97)
2. [It got later converted to a log message and got INFO level (no more debug) but was unseen](https://github.com/Irqbalance/irqbalance/commit/c66ee249bdc356a39f81d389ec528730a1ab0451#diff-74a30482d921ab3bffa86b0b1d5d481f70bff90c7ace5089813c521b98be07efR321)
3. [it was moved around when introducing sockets, still unseen](https://github.com/Irqbalance/irqbalance/commit/d1993bcde2a524346a9508754671f096ec129ad1#diff-74a30482d921ab3bffa86b0b1d5d481f70bff90c7ace5089813c521b98be07efR255)
4. [recently not only foreground but also log_journal is auto-set](https://github.com/Irqbalance/irqbalance/commit/75a15c5dbe6b3a1fd50569b94359d22ae4116285)

Combined that leads to Ubuntu, Debian and probably all other users of the recent version with change number 4 to get continuous journal entries like this every 10 seconds:

```
Jan 05 12:30:55 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:05 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:15 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:25 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:35 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:45 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
```

It grabs a lot of space and it does not say what it is about.
That made sense back then when added as a message for the developer, but today is just annoying in the log.

[...]

PR #293, opened 2024-01-05
Stop log spam due to debug message from scan

cpaelzer:
Currently systems encounter log spam every 10 seconds due to a message meant for debugging now being forwarded to journald. Users get three empty lines followed by 78 "-" without any further info what this is about.

```
Jan 05 12:31:25 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:35 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
Jan 05 12:31:45 n-vm irqbalance[770]: 
                                      
                                      
                                      -----------------------------------------------------------------------------
```

Due to its original debug-only purpose the message also grabs a lot of screen-estate. So while touching it anyway let us also make it a bit less agressive (no three lead line breaks, less width), now it is like:
```
----- Enter: scan -----
```

Fixes: #292

I was just seeing this and tried to provide a simple fix, I do not insist on the wording, the style or anything.
All I want is to get the log spam stopped, so please feel free to modify it along the way if you'd want it to look different.

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -289,7 +289,7 @@ gboolean force_rescan(gpointer data __attribute__((unused)))
 
 gboolean scan(gpointer data __attribute__((unused)))
 {
-	log(TO_CONSOLE, LOG_INFO, "\n\n\n-----------------------------------------------------------------------------\n");
+	log(TO_CONSOLE, LOG_DEBUG, "----- Enter: %s -----\n", __func__);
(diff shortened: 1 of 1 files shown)


cpaelzer:
This was found to be a dup of the fix done by f3282f4d - closing

PR #299, opened 2024-03-13
irqbalance-ui: check if using a negative index of buffer

liutgnu:
A negative index will be used when recv() fails, which is unexpected for the data buffer.


liutgnu:
Hi @nhorman,

The issue was found by Static Application Security Testing (SAST), which is a potential weakness IMHO. Could you please have a review to see if it should be fixed?

Thanks!

liutgnu:
> yeah, looks legitimate. Send a PR?

Hi nhorman,

The PR link is: https://github.com/Irqbalance/irqbalance/pull/299/files. Not sure if this is what you want.

Thanks!

PR #300, opened 2024-03-13
Fix socket API being blocked for 10s

### LINKED ISSUE #298: socket API blocks for 10s (sleep_interval)
In `scan()`, we call `sleep_approx(sleep_interval)`, blocking the socket API (blocking the main_loop)
Looking at the code
https://github.com/Irqbalance/irqbalance/blob/64285577b465c4c5d8b5bbfa0120432e1d5ffd63/irqbalance.c#L320-L322
could we just replace those 3 lines with a return and still reenter `scan()` sleep_interval later (called by `g_timeout_add_seconds`)

champtar:
Instead of sleeping in scan() and blocking the main loop, return and let the main loop call back scan().

Quickly tested (long running daemon and one shot), it seems to works
Fix #298

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -75,20 +75,6 @@ char socket_name[64];
 char *banned_cpumask_from_ui = NULL;
 #endif
 
-static void sleep_approx(int seconds)
-{
-	struct timespec ts;
-	struct timeval tv;
-	gettimeofday(&tv, NULL);
-	ts.tv_sec = seconds;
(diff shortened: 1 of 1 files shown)


champtar:
Thanks @nhorman ! Do you plan to release 1.9.4 in a near-ish future ? (else I'll just backport this patch)

nhorman:
I can do a release next week, there appears to be sufficient content for one

PR #304, opened 2024-03-30
Check `fflush()` return value

balrog-kun:
Since `fprintf()` may buffer output, as noted in 470a64b19062, `fclose()`'s error value was also being checked for the write errors.  However in 8d7c78304fb9 an `fflush()` was added in between meaning that these buffered write errors were again unchecked.  Some actual errors were not being logged, in my case `-ENOSPCs`.

Make the fclose and fflush branches look similar.

Fixes: 8d7c78304fb9 ("Flush file before closing")


nhorman:
@cloudaeye review-code

deepakmeenax:
@cloudaeye review-code

PR #306, opened 2024-03-31
small cleanups

neheb:


neheb:
No idea if it makes more sense to change the type from void to char instead.

Issue #321, opened 2024-07-10
IRQBALANCE_ARGS should be explicitly defined to avoid systemd warning


Camelron:
Hello,

I noticed when updating systemd that some new noise was being produced by the systemd unit:

```
irqbalance: "Referenced but unset environment variable evaluates to an empty string: IRQBALANCE_ARGS"
```

It seems like systemd now logs when a variable expanded in a unit file is not explicitly defined. Thus, in the systemd service when we expand the variable this warning is logged https://github.com/Irqbalance/irqbalance/blob/master/misc/irqbalance.service#L11.  This behavior was introduced to systemd in versions 254+ https://github.com/systemd/systemd-stable/commit/f331434d13488425ccd8485327085d15f8f92aea 

I was able to squelch this warning by uncommenting IRQBALANCE_ARGS in irqbalance.env, instead setting it to the empty string. Might be worth handling upstream?

nhorman:
yeah, that seems reasonable, if you would like to open a PR, I'd accept it.

Though that said, the systemd change seems somewhat...agressive.  lots of services have optional environment variables...

PR #322, opened 2024-07-11
Define IRQBALANCE_ARGS as empty string to squelch systemd warning

Camelron:
Systemd now logs when a variable expanded in a unit file is not explicitly defined. Thus, in the systemd service when we expand the variable a warning is logged https://github.com/Irqbalance/irqbalance/blob/master/misc/irqbalance.service#L11. This behavior was introduced to systemd in versions 254+ https://github.com/systemd/systemd-stable/commit/f331434d13488425ccd8485327085d15f8f92aea

Squelch this warning by uncommenting IRQBALANCE_ARGS in irqbalance.env, instead setting it to the empty string.

Addresses https://github.com/Irqbalance/irqbalance/issues/321

### DIFF
--- a/misc/irqbalance.env
+++ b/misc/irqbalance.env
@@ -41,4 +41,4 @@
 #    Append any args here to the irqbalance daemon as documented in the man
 #    page.
 #
-#IRQBALANCE_ARGS=
+IRQBALANCE_ARGS=""


Camelron:
Should we also do the same with all of the other IRQBALANCE_* variables in irqbalance.env?

nhorman:
I guess it wouldn't hurt, but its really making me question the validity of what systemd is doing here.  There are many cases in which environment variables are optional - most code in fact does the check:
```
var = getnev("ENVVAR");
if (!var)
   var = <default value>
```

So I really don't see how systemd can consider it an error.  Informational, or a warning perhaps, but certainly not an error.

I've opened https://github.com/systemd/systemd-stable/issues/425 to enquire about this

PR #334, opened 2024-11-20
Advance the signal handler add

601376534:
When there are too many devices in the /sys/bus/pci/devices directory, it will increase the service startup time. In my environment, the build_dev_irqs() function takes more than 30 seconds. During this time, the service will not be able to respond to some signals (such as the -15 signal sent by systemd).

### DIFF
--- a/irqbalance.c
+++ b/irqbalance.c
@@ -664,6 +664,13 @@ int main(int argc, char** argv)
 		}
 	}
 
+	g_unix_signal_add(SIGINT, handler, NULL);
+	g_unix_signal_add(SIGTERM, handler, NULL);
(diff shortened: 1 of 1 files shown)


601376534:
Sorry to get back to you so late.
In my environment, the environment information (such as banned_cpus) of irqbalance may be modified frequently and irregularly,Therefore, i need to restart the irqbalance service to modify the banned CPUs list in a timely manner, When the startup is too slow,After the irqbalance cannot receive signals, I guess the last modification may not take effect, causing other problems in my env.

I went to understand the process of the systemctl restart command, and I agree with you that it is incorrect to advance the signal response, which goes against the original purpose of the systemctl restart command.

PR #337, opened 2024-12-25
fix some GCC fanalyzer warnings

neheb:
-Wfanalyzer complains about null pointer dereferences. Check for them.

Interestingly I see do loops in quite a few places.


neheb:
Ah yeah: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-do-while.html

PR #354, opened 2025-11-20
meson: force systemd-service installation with a seperate option

fossdd:
Allows the systemd service installation without systemd build-dependency (as we in Alpine allow systemd services to be included for downstreams, but don't have systemd in Alpine).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -76,7 +76,7 @@ executable(
 
 install_man('irqbalance.1')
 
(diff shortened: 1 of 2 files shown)


fossdd:
Would be awesome for downstream to also switch to the meson build system and include the service, but no pressure! Thanks for maintaining irqbalance and the fast responses

PR #356, opened 2026-02-25
docs: document an edge case for `IRQBALANCE_BANNED_CPULIST`

Tal-or:
In a case when user wants to set the `IRQBALANCE_BANNED_CPULIST` variable as empty,
it cannot use the empty string `""` for that, since irqbalance apply a custom logic for not placing  IRQs on `isolated` or `nohz_full` cpus by default.

This PR adds a documentation for how to override the default behaviour and set an empty banned list where all CPUs are participating in the IRQ balancing.  

Signed-off-by: Talor Itzhak <titzhak@redhat.com>


liuchao173:
You can use the empty string "" now after https://github.com/Irqbalance/irqbalance/pull/358 rather than an invalid value

Issue #355, opened 2026-02-17
PR #312 ENOSPC Mitigation Incomplete: Placement Algorithm Keeps Targeting Saturated CPU


suresh-thelkar:
## Summary

PR #312 introduced ENOSPC mitigation to prevent persistent "Cannot change IRQ affinity: No space left on device" errors when a CPU's vector table is full. However, our testing reveals that the mitigation does not achieve the expected behavior of avoiding ENOSPC-triggering CPUs in subsequent iterations in the scenarios we mentioned below. The placement algorithm keeps targeting the same saturated CPU, resulting in continuous ENOSPC errors.

## Environment

- **System**: Server with 128 CPUs, kernel configured with isolated CPUs (`isolcpus=4-63,68-127`)
- **Kernel**: Linux 6.6.x, Azure Linux 3.0([PR to upgrade to 1.9.5](https://github.com/microsoft/azurelinux/pull/15379))
- **irqbalance versions tested**: 1.9.3 (without PR #312) and 1.9.5 (with PR #312)

## Expected Behavior (from PR #312)

After irqbalance encounters ENOSPC trying to place IRQs onto a CPU (CPU is "full"):
1. Subsequent placement attempts should avoid targeting the same saturated CPU
2. With multiple eligible CPUs still available, scheduling should shift toward those CPUs
3. The persistent "Cannot change IRQ N affinity: No space left on device" spam should be reduced

## Test Case Setup

To reproduce the ENOSPC scenario in a controlled manner:

1. **Constrain eligible CPUs**: Use `IRQBALANCE_BANNED_CPULIST=3-127` to allow only CPUs 0, 1, 2
2. **Enable debug mode**: `irqbalance --foreground --debug -t 1` (1-second rebalance interval)
3. **Monitor affinity writes**: `strace -ff -ttt -e write -y` to capture `/proc/irq/*/smp_affinity` operations
4. **Track vector allocation**: Monitor `/proc/irq/*/effective_affinity` for actual kernel vector distribution

The constrained 3-CPU scenario causes CPU vector slots to fill up quickly, triggering the ENOSPC condition reliably.

### 3-CPU Constraint Test (60 seconds)

| Metric | irqbalance 1.9.3 (no PR #312) | irqbalance 1.9.5 (with PR #312) |
|--------|-------------------------------|----------------------------------|
| ENOSPC count | 1,752 | 1,854 |
[...]

PR #360, opened 2026-04-13
Fix ENOSPC handling with slot-aware placement and fallback

suresh-thelkar:
### Summary

This PR fixes incomplete ENOSPC handling when CPU interrupt vector tables become saturated. The fix adds proactive avoidance of saturated CPUs during placement and reactive fallback when ENOSPC occurs.

**Key design principle:** This patch does NOT modify the core load-balancing algorithm. All changes are additive and only affect behavior when ENOSPC conditions are detected. Normal operation remains completely unchanged.

### Problem

When systems have constrained CPU sets (e.g., `isolcpus` or `IRQBALANCE_BANNED_CPULIST`), CPUs can reach their interrupt vector capacity limit (~224 vectors). The kernel returns `ENOSPC` when attempting to assign more IRQs to a full CPU.

Current behavior:
- irqbalance keeps targeting the same saturated CPU
- ENOSPC errors occur at ~35/s indefinitely
- No adaptation or fallback mechanism

More details including the various test scenarios can be seen in the issue- https://github.com/Irqbalance/irqbalance/issues/355 that we have created

### Solution

This patch builds upon the foundation laid by PR #312, which introduced `slots_left` tracking but left gaps in the placement algorithm and recovery path.

**1. Slot-Aware Placement (`placement.c`)** — *Minimal, targeted enhancement*

```c
/* Skip fully saturated CPUs — only when slots_left <= 0 */
if (d->slots_left <= 0)
    return;

/* Add penalty ONLY for low-capacity CPUs (slots_left < 10) */
/* When slots_left >= 10 or INT_MAX: penalty = 0, load-based selection unchanged */
if (d->slots_left < SLOTS_PENALTY_THRESHOLD)
    slots_penalty = (SLOTS_PENALTY_THRESHOLD - d->slots_left) * SLOTS_PENALTY_FACTOR;

/* Use slots_left as tie-breaker ONLY when loads are equal */
if (adjusted_cost == best_adjusted_cost) {
    if (d->slots_left > best->best->slots_left)
        best->best = d;
}
```

**Why this doesn't change core logic:**
- At cycle start, `slots_left = INT_MAX` for all CPUs → penalty = 0
- Load remains the primary (and usually only) selection factor
- Penalty only kicks in after ENOSPC has been observed

**2. ENOSPC Fallback Mechanism (`activate.c`)** — *Error path only*

This code is ONLY invoked when `activate_mapping()` receives ENOSPC from the kernel:

1. Mark original CPU as saturated (`slots_left = 0`)
2. Try alternative CPUs in priority order:
   - Same cache domain (best locality)
   - Same NUMA node (recursive traversal)
   - Any available CPU
3. Up to 8 attempts with `tried_cpus` bitmask. Why 8, please look at the section "Macro Constants Rationale"
4. Use existing `migrate_irq_obj()` for proper list management

**Why this is safe:**
- `try_fallback_cpu()` is only called from the ENOSPC error handler
- Never invoked during normal successful placements
- Respects existing topology and banned CPU constraints

### Test Results

**Environment:**
- Hardware: Intel Xeon Gold 6438N, 128 CPUs
- Kernel: 6.6.x
- `isolcpus=4-63,68-127` (8 non-isolated CPUs: 0-3, 64-67)
- Test config: `IRQBALANCE_BANNED_CPULIST=3-127`, rebalance interval=1s
- Result: 3 CPUs available for IRQ placement (0, 1, 2), 586 IRQs

**30-Minute Test Results:**

| Metric | Before (1.9.3) | After (1.9.5 with the changes in this PR) | Change |
|--------|----------------|---------------|--------|
| ENOSPC errors | 63,135 | 7,801 | **-87.6%** |
| OK writes CPU 0 | 543 | 1,101 | **+103%** |
| OK writes CPU 1 | 666 | 1,758 | **+164%** |
| OK writes CPU 2 | 694 | 1,083 | **+56%** |

image

image

image

**5-Minute Debug Mode Validation:**

The last screenshot above show irqbalance running with `--debug` flag, which outputs detailed placement and fallback decisions.

**Key observations**:

1. **Proactive avoidance**: Slot-aware penalty in `find_best_object()` steers new IRQs away from CPUs with low `slots_left`, reducing ENOSPC occurrences at the source
2. **Reactive fallback**: When ENOSPC does occur, `try_fallback_cpu()` immediately retries on alternative CPUs (cache siblings → NUMA siblings → any available), with 100% success rate in this test
3. **Load distribution**: IRQs are distributed across all 3 available CPUs (0, 1, 2) rather than clustering on a single saturated CPU

This validates that both mechanisms work in tandem — proactive placement reduces the frequency of ENOSPC, while reactive fallback handles the remaining cases without deferring to the next cycle. 

**Fallback Mechanism Validation:**
```
=== Fallback Attempts ===
IRQ 226: ENOSPC fallback from CPU 1 to CPU 2 (attempt 1, slots_left=2147483647)
IRQ 226: successfully placed on fallback CPU 2 (slots_left now 2147483647)
...
IRQ 664: ENOSPC fallback from CPU 1 to CPU 2 (attempt 1, slots_left=2147483647)
IRQ 664: fallback CPU 2 saturated (ENOSPC), slots_left set to 0
IRQ 664: ENOSPC fallback from CPU 1 to CPU 0 (attempt 2, slots_left=2147483647)
IRQ 664: successfully placed on fallback CPU 0 (slots_left now 2147483647)

=== Successful Fallbacks ===
62 IRQs successfully placed via fallback

=== Failed Fallbacks ===
(none)
```

### Files Changed

- `activate.c`: fallback mechanism, safety improvements
- `placement.c`: slot-aware penalty, traversal fix

### Design Principles — Why This Is Safe for Upstream

| Principle | How This Patch Adheres |
|-----------|------------------------|
| **Core algorithm unchanged** | Load-based placement remains primary; `slots_penalty` is zero under normal conditions (`slots_left >= 10` or `INT_MAX`) |
| **Error path only** | `try_fallback_cpu()` only invoked on ENOSPC; normal placements never trigger fallback |
| **Builds on existing work** | Extends PR #312's `slots_left` tracking rather than replacing it |
| **Defensive additions** | NULL checks, `snprintf()`, errno capture — purely protective, no behavior change |
| **Bounded state** | `slots_left` clamped to 0 instead of drifting negative; predictable recovery |
| **Existing APIs** | Uses `migrate_irq_obj()` for proper interrupt list management |

### Backward Compatibility

- No changes to command-line interface
- No changes to configuration options  
- No changes to config file format
- **Zero impact on normal operation** — penalty is 0 when `slots_left >= 10` (always true at cycle start)
- Only affects ENOSPC error handling path
- Tested on systems without CPU constraints also — no regressions observed

### Macro Constants Rationale

| Constant | Value | Rationale |
|----------|-------|-----------|
| `SLOTS_PENALTY_THRESHOLD` | 10 | CPUs with ≥10 slots remaining have ample headroom; penalty is zero, so pure load-based selection applies. Below 10, we start preferring CPUs with more capacity to reduce ENOSPC likelihood. |
| `SLOTS_PENALTY_FACTOR` | 1000 | Typical IRQ load values range from thousands to millions. A penalty of 1000–9000 (for slots 9→1) is significant enough to influence placement when loads are similar, but not so large as to override load decisions entirely. Example: CPU with load=5000, slots=2 → cost=13000; CPU with load=8000, slots=10 → cost=8000 (preferred). |
| `MAX_FALLBACK_ATTEMPTS` | 8 | Based on typical x86 cache topology: 2–8 physical cores share an L3 cache domain. 8 attempts is sufficient to exhaust a cache domain before escalating to NUMA/global scope. The loop exits early when no valid candidates remain (all `slots_left ≤ 0`), so this limit is rarely hit in practice. |

### References

- Issue: https://github.com/Irqbalance/irqbalance/issues/355
- Related PR: https://github.com/Irqbalance/irqbalance/pull/312 (original ENOSPC mitigation)

### DIFF
--- a/activate.c
+++ b/activate.c
@@ -90,33 +90,6 @@ static int is_valid_fallback(struct topo_obj *cpu, struct topo_obj *original,
 }
 
 /*
- * Find the CPU with the most available slots among valid candidates.
- * This avoids the "first match bias" that can cause IRQs to cluster
- * on the same fallback CPU repeatedly.
- *
- * Returns the best CPU or NULL if none found.
- */
-static struct topo_obj *find_best_fallback_in_list(GList *list,
-						   struct topo_obj *original,
-						   cpumask_t *tried_cpus)
-{
-	struct topo_obj *best = NULL;
-	GList *iter;
-
-	for (iter = list; iter; iter = iter->next) {
-		struct topo_obj *cpu = iter->data;
-
-		if (!is_valid_fallback(cpu, original, tried_cpus))
-			continue;
-
-		/* FIX: Select CPU with most slots_left to avoid clustering */
-		if (!best || cpu->slots_left > best->slots_left)
-			best = cpu;
-	}
-	return best;
-}
-
-/*
  * Recursively collect all CPU objects under a given topology object.
  * This fixes the NUMA fallback issue where numa_node->children contains
  * packages/caches, not CPUs directly.
@@ -160,15 +133,41 @@ static struct topo_obj *find_best_cpu_under_obj(struct topo_obj *obj,
 }
 
 /*
+ * Convert balance level to string for logging.
+ */
+static const char *balance_level_str(int level)
+{
+	switch (level) {
+	case BALANCE_NONE:
+		return "none";
+	case BALANCE_PACKAGE:
+		return "package";
+	case BALANCE_CACHE:
+		return "cache";
+	case BALANCE_CORE:
+		return "core";
+	default:
+		return "unknown";
+	}
+}
+
+/*
  * Try to find an alternative CPU when the primary target returns ENOSPC.
  * Returns 0 on success (IRQ placed on alternative CPU), -1 on failure.
  *
  * Uses iteration with a tried_cpus bitmask to avoid retrying the same CPU.
  *
- * Priority order:
- *   1. Same cache domain CPUs with slots_left > 0 (best slots_left first)
- *   2. Same NUMA node CPUs with slots_left > 0 (traverses full tree)
- *   3. Any allowed CPU with slots_left > 0 (best slots_left first)
+ * IMPORTANT: This function respects the IRQ's configured balance level.
+ * If the user configured an IRQ to balance at a specific level (e.g., cache),
+ * fallback will only search within that scope. If no CPU is available within
+ * the constrained scope, we warn the user and fail rather than violating the
+ * configured policy.
+ *
+ * Scope by balance level:
+ *   - BALANCE_CORE: No fallback possible (single CPU, fail immediately)
+ *   - BALANCE_CACHE: Only search within same cache domain
+ *   - BALANCE_PACKAGE: Only search within same package
+ *   - BALANCE_NONE: Search within same NUMA node (widest scope)
  *
  * FIXES APPLIED:
  *   1. NUMA fallback now traverses topology tree to find actual CPUs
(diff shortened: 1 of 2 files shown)


suresh-thelkar:
@nhorman, As requested in [the issue ](https://github.com/Irqbalance/irqbalance/issues/355#issuecomment-4233810906), When you have a moment, could you please take a look into this PR and let me know if you have any feedback or concerns. Happy to address any review comments.

activate.c
```diff
@@ -34,6 +34,297 @@
...
+	int saved_errno;
+	int attempts = 0;
+
+	cpus_clear(tried_cpus);
+
+	/* Find cache domain parent */
+	cache_parent = original->parent;
+	while (cache_parent && cache_parent->obj_type != OBJ_TYPE_CACHE)
+		cache_parent = cache_parent->parent;
+
```
nhorman, line 251: how is this honoring any potential balance level that the user may have assigned to the irq?  i.e. if the user configured a policy that expressly said "balance this irq at the cache level", you can't just ignore that and go up the object tree to find something that works.   If you can't find a workable cpu within the target range that the irq is set to balance on, you need to warn the user and fail.

nhorman (changes requested):

activate.c
suresh-thelkar: @nhorman, Thank you very much for the review feedback — you were absolutely right.

I'm relatively new to the irqbalance codebase and still getting my hands dirty with the internals. Honestly, I was not aware of the `--policyscript` feature with `balance_level` settings and we didn't test with it. Thanks for cathing this: the original code would escalate beyond the configured scope (cache → NUMA → any CPU), violating user policy if configured.

I've pushed a fix that addresses this:

**Key changes:**
- `BALANCE_CORE`: No fallback — warn and fail immediately  
- `BALANCE_CACHE/PACKAGE/NONE`: Search only within the configured scope; warn and fail if exhausted
- No escalation beyond user-configured policy

Could you please review([the latest commit](https://github.com/Irqbalance/irqbalance/pull/360/changes/4332ecbfdef49fa901354e05e1d5c82cd1f76870)) and let me know if this addresses your concern?

nhorman (approved):
yeah, that looks reasonable

